PR #1, opened 2021-07-02
Updating server operator and removing worker

knkski:
Updates the server operator to be more in line with existing Kubeflow operators on a number of fronts, such as using the SDI library, and adds Github Actions-based CI.


kwmonroe (approved):
LGTM.  I especially like the attention to tests. nice job!

PR #2, opened 2021-07-16
Remove prometheus relation

knkski:
It uses an older Prometheus charm that is deprecated in favor of

https://charmhub.io/prometheus-k8s


joedborg (approved):
LGTM

PR #13, opened 2022-02-15
fix: add trailing '/' to uri in ingress

ca-scribner:
fixes canonical/bundle-kubeflow#436

This fixes the istio ingress/virtualservice routing.  

Previously, the rewrite rule of `prefix=/mlflow` would serve a blank page at `INGRESS/mlflow` and a dead link at `INGRESS/mlflow/.  This appeared to be because the rewrite rules would break the static files that mlflow serves.  With this old rule (no trailing slash in the prefix) the static files get served at `INGRESS/static-files/...` rather than `INGRESS/mlflow/static-files/...` and are thus inaccessible.  

With this PR, the rewrite rule of `prefix=mlflow/` (including trailing slash) serves the correct mlflow UI at `INGRESS/mlflow/` with the static files correctly being accessed from `INGRESS/mlflow/static-files/...`.  But, this serves nothing at `INGRESS/mlflow`, which might be counter intuitive.  I tried to find some extra istio virtualiservice config that would get both to work, but I was not successful.

# Test instructions:
On a deployment with at least the istio-pilot, gateway, and possibly kubeflow-dashboard, do:
```
juju deploy mlflow-server
juju deploy cs:~charmed-osm/mariadb-k8s-35 mysql-mlflow
juju deploy minio minio-mlflow

juju relate mlflow-server mysql-mlflow
juju relate mlflow-server:object-storage minio-mlflow:object-storage
juju relate mlflow-server:ingress istio-pilot:ingress
```

Then browse to `INGRESSIP/mlflow/` (including trailing `/`).  The auth flow should occur and then get you to the mlflow UI

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -61,8 +61,11 @@ jobs:
         set -eux
         juju model-config update-status-hook-interval=15s
         juju deploy ch:istio --trust
-        juju config istio-pilot 'default-gateways=kubeflow-gateway'
-        sleep 10
+        juju deploy istio-gateway --channel=1.5/stable --trust istio-ingressgateway
+        juju deploy istio-pilot --channel=1.5/stable --config default-gateway=kubeflow-gateway
+        juju relate istio-pilot:istio-pilot istio-ingressgateway:istio-pilot
+        sleep 30
+        kubectl patch role -n kubeflow istio-ingressgateway-operator -p '{"apiVersion":"rbac.authorization.k8s.io/v1","kind":"Role","metadata":{"name":"istio-ingressgateway-operator"},"rules":[{"apiGroups":["*"],"resources":["*"],"verbs":["*"]}]}'
         juju wait -wvt 300 --retry_errors 20
         juju model-config update-status-hook-interval=5m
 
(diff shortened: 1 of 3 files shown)


johnsca (commented):
I know the trailing slash isn't ideal, but given how difficult it seems to be to get the Istio prefix rules to work how it seems like they should, this seems like a reasonable workaround. However, I'm confused as to why this issue wasn't caught by the integration test, considering it does use Istio and the ingress gateway address. Should the test be improved to catch this and avoid a regression?

DomFleischmann:
> I know the trailing slash isn't ideal, but given how difficult it seems to be to get the Istio prefix rules to work how it seems like they should, this seems like a reasonable workaround. However, I'm confused as to why this issue wasn't caught by the integration test, considering it does use Istio and the ingress gateway address. Should the test be improved to catch this and avoid a regression?

Yes I agree with @johnsca looking at the integration test, [this line](https://github.com/canonical/mlflow-operator/blob/c9997ad43c08e0c426ad49304515906a7dc11286/tests/test_selenium.py#L63) might not be enough to actually ensure we have a working dashboard, one additional check would probably be enough for now.

ca-scribner:
Good point on adding more tests.  I get the feeling from [this line](https://github.com/canonical/mlflow-operator/blob/c9997ad43c08e0c426ad49304515906a7dc11286/tests/test_selenium.py#L62) that it might be a good idea, too

DomFleischmann (approved):

PR #22, opened 2022-02-25
Store secrets in kubernetes secrets

agathanatasha:
- pod env reference the existing object storage kubernetes secrets instead of plain text
- put db password and tracking uri in kubernetes

### DIFF
--- a/charms/mlflow-server/src/charm.py
+++ b/charms/mlflow-server/src/charm.py
@@ -57,13 +57,15 @@ class Operator(CharmBase):
         endpoint = f"http://{obj_storage['service']}:{obj_storage['port']}"
         tracking = f"{self.model.app.name}.{self.model.name}.svc.cluster.local"
         tracking = f"http://{tracking}:{config['mlflow-port']}"
-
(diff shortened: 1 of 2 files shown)


ca-scribner (commented):
Not something you added, but noticed this charm has a bug we noticed and fixed elsewhere in the `CheckFailedError`.  See [here](https://github.com/canonical/kubeflow-profiles-operator/blob/1a9948e48faa4da6d6d15ac1656f47442fa7f8a7/src/charm.py#L202), which forces `self.msg = str(msg)` (the status constructors raise if their message is not a string, so we force it here to make sure it works)

ca-scribner (approved):
Apart from fixing the minor bug in the previous comment, this LGTM.  I'm not 100% sure whether the S3 storage is working properly, but I'm also not sure it was working prior to this change and I need to look into something similar in bundle-kubeflow#429 anyway so I'll raise an issue if something needs fixing.

agathanatasha:
@ca-scribner I have updated the error message

PR #27, opened 2022-03-02
feat: Add a secret to use with seldon deployments

### LINKED ISSUE #26: Add/update `seldon-init-container-secret` for seldon
Discussed in more detail [here](https://github.com/canonical/bundle-kubeflow/issues/429)

The `seldon-init-container-secret` can be used by `SeldonDeployment`s to provide credentials for pulling models out of s3 storage.  This secret was previously implemented in the mlflow charm as a convenience to make using seldon deployments easier, but the naming convention of the secrets did not match the current seldon init-containers expectations and it had been removed in #22.

Proposed here is that we add the `seldon-init-container-secret` back in with updated formatting to match Seldon's expectations (see [seldondocs ](https://docs.seldon.io/projects/seldon-core/en/latest/servers/kfserving-storage-initializer.html#create-a-secret-containing-the-environment-variables) or [this issue](https://github.com/canonical/bundle-kubeflow/issues/429) for details).  Naming may be changed to assure uniqueness and reduce confusion (`seldon-init-container-secret` implies it might have been created by seldon, not for seldon)

ca-scribner:
Fixes #26.  Charm creates a secret with mlflow's s3 credentials called  `{charm_name}-seldon-init-container-s3-credentials`.  This secret is formatted as needed for a [Seldon Deployment Init Container](https://docs.seldon.io/projects/seldon-core/en/latest/servers/kfserving-storage-initializer.html#create-a-secret-containing-the-environment-variables).

Also:
* adds tests for the new secret
* refactors some secret creation to make the main function of the charm read easier
* removes an unused dependency from requirements.txt

### DIFF
--- a/charms/mlflow-server/src/charm.py
+++ b/charms/mlflow-server/src/charm.py
@@ -320,7 +320,7 @@ def _db_secret_dict(mysql):
     db_secret = {
         "DB_ROOT_PASSWORD": mysql["root_password"],
         "MLFLOW_TRACKING_URI": f"mysql+pymysql://root:{mysql['root_password']}@{mysql['host']}"
-                               f":{mysql['port']}/{mysql['database']}",
+        f":{mysql['port']}/{mysql['database']}",
     }
     return _b64_encode_dict(db_secret)
 
--- a/charms/mlflow-server/src/charm.py
+++ b/charms/mlflow-server/src/charm.py
@@ -131,32 +131,13 @@ class Operator(CharmBase):
         secrets = [
             {
                 "name": f"{charm_name}-minio-secret",
-                "data": {
-                    k: b64encode(v.encode("utf-8")).decode("utf-8")
-                    for k, v in {
-                        "AWS_ENDPOINT_URL": "http://{service}:{port}".format(**obj_storage),
-                        "AWS_ACCESS_KEY_ID": obj_storage["access-key"],
-                        "AWS_SECRET_ACCESS_KEY": obj_storage["secret-key"],
-                        "USE_SSL": str(obj_storage["secure"]).lower(),
-                    }.items()
-                },
+                "data": _minio_credentials_dict(obj_storage=obj_storage),
             },
             {
-                "name": f"{charm_name}-db-secret",
-                "data": {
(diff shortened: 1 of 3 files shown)


charms/mlflow-server/src/charm.py
```diff
@@ -287,5 +268,45 @@ def __init__(self, msg, status_type=None):
...
+    minio_credentials = {
+        "AWS_ENDPOINT_URL": f"http://{obj_storage['service']}:{obj_storage['port']}",
+        "AWS_ACCESS_KEY_ID": obj_storage["access-key"],
+        "AWS_SECRET_ACCESS_KEY": obj_storage["secret-key"],
+        "USE_SSL": str(obj_storage["secure"]).lower(),
+    }
+    return _b64_encode_dict(minio_credentials)
+
+
+def _seldon_credentials_dict(obj_storage):
```
DomFleischmann, line 288: how does do the seldon-init containers get these credentials?  Specially when the seldon deployment is in a different namespace.

DomFleischmann (commented):
Looks very promising, I'm just a bit confused about all of this will connect.

charms/mlflow-server/src/charm.py
ca-scribner: For now, only with some extra steps :(  There's more context in [this comment](https://github.com/canonical/bundle-kubeflow/issues/429#issuecomment-1055755776).  This is more a stop-gap fix until we have a way of injecting secrets into everyone's namespaces. 

This first step was decided on with @Barteus so that:
* for deployments with Seldon and MLFlow in the same namespace, `SeldonDeployment`s work easily by referencing this secret directly
* for deployments with Seldon and MLFlow in different namespaces, `SeldonDeployment`s **cannot** reference this secret directly but at least there's a "template" secret in mlflow's namespace that is easy to copy as needed

As soon as we can inject secrets into user namespaces, we remove this secret.

After talking with @Barteus, this secret means if seldon

DomFleischmann (approved):

DomFleischmann:
Looks good, can you resolve the conflict @ca-scribner ?

PR #28, opened 2022-03-08
feat: add mlflow s3 bucket name as config

### LINKED ISSUE #24: MLFlow hardcoded bucket name - impossible to use MLFlow with AWS S3
https://github.com/canonical/mlflow-operator/blob/c856446074868d4735627c95878960d91555f4da/charms/mlflow-server/src/charm.py#L20

The name of the bucket for MLFlow is hardcoded. This is a big issue because this makes using Minio in Gateway mode + MLFlow impossible on AWS (S3 buckets are globally unique).

It's a good first issue :)

ca-scribner:
Adds the "default_artifact_root" config value, which sets the s3 bucket that Argo Workflows uses as the default storage bucket (where it puts data if a Workflow does not specify its own bucket).  This is also useful in gateway mode, where you may use a bucket in an s3 store that already exists with a name.

fixes #24

### DIFF
--- a/charms/mlflow-server/src/charm.py
+++ b/charms/mlflow-server/src/charm.py
@@ -9,6 +9,7 @@ https://github.com/canonical/mlflow-operator
 
 import json
 import logging
+import re
 from base64 import b64encode
 
 from oci_image import OCIImageResource, OCIImageResourceError
@@ -22,7 +23,6 @@ from serialized_data_interface import (
 )
 
 DB_NAME = "mlflow"
-BUCKET_NAME = "mlflow"
 
 
 class Operator(CharmBase):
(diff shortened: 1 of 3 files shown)


charms/mlflow-server/src/charm.py
```diff
@@ -276,6 +277,22 @@ def _check_image_details(self):
         return image_details
 
 
+def validate_s3_bucket_name(name):
+    """Validates the name as a valid S3 bucket name, raising a CheckFailedError if invalid."""
+    # regex from https://stackoverflow.com/a/50484916/5394584
+    if re.match(
+        r"(?=^.{3,63}$)(?!^(\d+\.)+\d+$)(^(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])$)",
```
DnPlas, line 284: I'm not a big fan of `re`, but since we are doing this, we could use `re.compile` here as we are going to be calling this many times during the charm's lifecycle.

DnPlas (commented):
Just a minor comment on `re`, otherwise LGTM.

Issue #15, opened 2022-02-16
Expand test coverage


ca-scribner:
Current integration tests only check if an mlflow frontpage gets served.  We should add more coverage such as workload tests that actually put something to mlflow/try to retrieve it.

ca-scribner:
What should be in/out of scope for testing?  Should we test direct to mlflow, or should we test through an ingress?  And how do we incorporate these tests into the context of the full bundle (in the full bundle, tests going through ingress will also need authentication)?

ca-scribner:
copied from #17 which was almost exactly a duplicate:

We do not have test coverage for UI access with authentication

PR #29, opened 2022-03-09
Fix up mlflow-server

agathanatasha:
- Fix issues that causes failure when relating the pod defaults interface
  1) KeyError when getting object storage interface data
  2) Typo for getting mlflow port config value
  3) Fail to load `mlflow-requirements.txt` because it wasn't packed in the charm
- Add namespace to MLFLOW_S3_ENDPOINT_URL, or else it would fail to log model following the example script in readme
- Removed unused scripts
- Updated readme
  - follow official quick start guide to deploy kubeflow
  - add step to copy pod defaults to user's namespace (temporary workaround)
  - create minio bucket in example model script if default bucket is not created 
- Added tests


README.md
```diff
@@ -50,6 +41,7 @@ import os
 import warnings
 import sys
 
+import boto3
```
DomFleischmann, line 44: I feel like this code example could be moved to an `examples/` folder in this repository. Having this in the Readme before the instructions to access the dashboard feels strange.

DomFleischmann (commented):
This looks great! Just some small comments.

tox.ini
```diff
@@ -3,16 +3,16 @@ max-line-length = 100
 
 [tox]
 skipsdist = True
-envlist = {server}
+envlist = {mlflow-server}-{unit,lint,integration},selenium
```
ca-scribner, line 6: Interestingly, the `mlflow` here doesn't actually matter.  You can still run the commands without it (try `tox -e server-lint`).

Also, I think `fmt` is missing from the second list.  We can still use `tox -e server-fmt`, but that env wont show up in `tox -l` output unless we add it to this line.

.github/workflows/integration.yaml
```diff
@@ -7,123 +7,58 @@ on:
...
+          - mlflow-server
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          provider: microk8s
+          channel: 1.21/stable
+          charmcraft-channel: latest/candidate
+      - run: sg microk8s -c 'microk8s enable metallb:10.64.140.43-10.64.140.49'
```
ca-scribner, line 52: Is this needed?  The actions-operator might do this for us already

README.md
```diff
@@ -1,35 +1,35 @@
 # Kubeflow + MLflow on Juju with Microk8s
 
-![screenshot](demo.png "Screenshot showing kubeflow notebook publishing to mlflow")
+## Get Started
```
ca-scribner, line 3: Should we offer both a with and without kubeflow option here?  Without kubeflow works too and is pretty simple, something like:
```
juju deploy mlflow-server
juju deploy charmed-osm-mariadb-k8s mlflow-db
juju deploy minio
juju relate minio mlflow-server
juju relate mlflow-db mlflow-server
```

At minimum, that'll be helpful for us debugging/developing

README.md
```diff
@@ -41,92 +41,8 @@ Now open the notebook server and paste the following code into two cells:
...
+Paste the [example model code](./examples/elastic_net_wine_model.ipynb) in a separate cell.
 
 Run both cells and observe that your model metrics are recorded in MLflow!
 
-To access MLflow, run:
-```
-microk8s kubectl get services -A|grep "mlflow-server"
-```
-And open the `mlflow` `ClusterIP` in the browser with `:5000` on the end.
+![screenshot](demo.png "Screenshot showing kubeflow notebook publishing to mlflow")
```
ca-scribner, line 48: Should we add an example showing people how to see their model in the artifact store?  Either by using the boto3 python package (could be added to the `.ipynb` above) or by using the `mc` client locally or in a k8s pod.  I don't think it is apparent to someone reading through this we've actually stored their model in a way that can be served later

charms/mlflow-server/charmcraft.yaml
```diff
@@ -1,13 +1,15 @@
...
     run-on:
-    - name: "ubuntu"
-      channel: "20.04"
+      - name: 'ubuntu'
+        channel: '20.04'
 parts:
   charm:
     charm-python-packages: [setuptools, pip]
+    prime:
+      - ./files
```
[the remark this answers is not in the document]
agathanatasha: The requirements are sent over to admission webhook when relating, but I don't know how it's being used after that

charms/mlflow-server/tests/integration/test_charm.py
```diff
@@ -0,0 +1,83 @@
...
+    status = await ops_test.model.get_status()
+    url = f"http://{status['applications'][istio_gateway]['public-address']}.nip.io/mlflow/"
+
+    for _ in range(60):
+        try:
+            requests.get(url, timeout=60)
+            break
+        except requests.ConnectionError:
+            sleep(5)
+    r = requests.get(url)
```
ca-scribner, line 82: We need a stronger test here than just getting a response.  An example of why is the problem solved in #13 (which, as I look back, I forgot to describe the symptoms of :man_facepalming:), where because of an ingress problem we only served a partial dashboard (got `200`, but did not get all the page assets).  That was fixed [here](https://github.com/canonical/mlflow-operator/commit/55531c9f08aba5111230cf9f50ae80f11262d7ea) by checking for a specific element on the page that came from those assets (not a perfect solution, but better than nothing)

charms/mlflow-server/tests/integration/test_charm.py
```diff
@@ -0,0 +1,83 @@
...
+    await ops_test.model.add_relation(CHARM_NAME, obj_storage)
+    await ops_test.model.add_relation(CHARM_NAME, db)
+    await ops_test.model.wait_for_idle(
+        status="active", raise_on_blocked=False, raise_on_error=False
+    )
+
+    assert ops_test.model.applications[CHARM_NAME].units[0].workload_status == "active"
+
+
+async def test_access_dashboard(ops_test: OpsTest, request):
```
ca-scribner, line 42: We can split this test up a bit more to make it clear what role everything plays.  2/3 of the code is really a fixture (set up istio ingress gateway).  At a minimum it would be helpful to denote this in comments, but I think we can address a few things at once here.  

This test checks more than just dashboard access, it checks dashboard access **through an ingress**.  We should test these two things separately (1: is dashboard accessible, 2: is dashboard accessible through the ingress).  That way if say istio breaks something we see the mlflow dashboard is ok but mlflow dashboard via ingress is broken, and vise versa if the dashboard actually breaks we don't debug istio

charms/mlflow-server/tests/integration/test_charm.py
```diff
@@ -0,0 +1,83 @@
...
+    await ops_test.model.deploy(obj_storage)
+
+    my_charm = await ops_test.build_charm(".")
+    image_path = METADATA["resources"]["oci-image"]["upstream-source"]
+    resources = {"oci-image": image_path}
+    await ops_test.model.deploy(my_charm, resources=resources)
+    await ops_test.model.add_relation(CHARM_NAME, obj_storage)
+    await ops_test.model.add_relation(CHARM_NAME, db)
+    await ops_test.model.wait_for_idle(
+        status="active", raise_on_blocked=False, raise_on_error=False
```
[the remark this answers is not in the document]
ca-scribner: Another way that might be clearer would be to have two fixtures, and pass them in as a parameter:

def istio_ingress():
  # deploy ingress
  yield dashboard_url
  # clean up ingress

def null_ingress():
  # does not deploy an ingress
  return dashboard_url

@pytest.mark.parametrize(
  "ingress",
  (
    (pytest.lazy_fixture('istio_ingress'),),
    (pytest.lazy_fixture('null_ingress'),),
  )
)
def test_access_dashboard(ingress):
  dashboard_url = ingress
  ...

This uses an extra dependency (pytest-lazy-fixture), but to me reads a bit clearer on what it is doing.  Curious to hear your thoughts!

charms/mlflow-server/tests/integration/test_charm.py
ca-scribner: Another way that might be clearer would be to have two fixtures, and pass them in as a parameter:

```
def istio_ingress():
  # deploy ingress
  yield dashboard_url
  # clean up ingress

def null_ingress():
  # does not deploy an ingress
  return dashboard_url

@pytest.mark.parametrize(
  "ingress",
  (
    (pytest.lazy_fixture('istio_ingress'),),
    (pytest.lazy_fixture('null_ingress'),),
  )
)
def test_access_dashboard(ingress):
  dashboard_url = ingress
  ...
```

This uses an extra dependency (pytest-lazy-fixture), but to me reads a bit clearer on what it is doing.  Curious to hear your thoughts!

charms/mlflow-server/tests/integration/test_charm.py
```diff
@@ -0,0 +1,83 @@
...
+    url = f"http://{status['applications'][istio_gateway]['public-address']}.nip.io/mlflow/"
+
+    for _ in range(60):
+        try:
+            requests.get(url, timeout=60)
+            break
+        except requests.ConnectionError:
+            sleep(5)
+    r = requests.get(url)
+    assert r.status_code == 200
```
ca-scribner, line 83: I don't know if this is in-scope or our-of-scope for this PR, but we our tests here only check the dashboard, not whether mlflow itself is working (can I store model details/artifacts and pull them out later).  We do have an example notebook to do this, but no tests.  That example could be added here as a workload test

If this is out of scope, that's ok just resolve this.  I think #15 captures this issue

charms/mlflow-server/tox.ini
```diff
@@ -54,4 +54,11 @@ commands =
 
 [testenv:unit]
 commands = 
-	pytest tests
+	pytest tests --ignore={[vars]tst_path}integration
```
ca-scribner, line 57: it is minor, but rather than say "unit tests are everything except `/tests/integration`" I think we should say "unit tests are `/tests/unit`".  It feels closer to the message our directory structure suggests.  
```suggestion
	pytest {[vars]tst_path}unit
```
Note that I haven't tested this command locally, just run off memory.  It'll show in the CI pretty quick though...

charms/mlflow-server/tox.ini
```diff
@@ -54,4 +54,11 @@ commands =
...
+
+[testenv:integration]
+deps = 
+    pytest-operator
+    lightkube
+    requests
+commands = pytest -v --tb native --asyncio-mode=auto --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
```
ca-scribner, line 64: Same comment here.  Let's say "integration tests are X" and not "integration tests are everything except Y"

.github/workflows/integration.yaml
agathanatasha: From what I can see [here](https://github.com/charmed-kubernetes/actions-operator/blob/main/src/bootstrap/index.ts#L33), it only enables dns storage and rbac

charms/mlflow-server/tox.ini
agathanatasha: will update this. I personally prefers stating the positives too, but [operator-template](https://github.com/canonical/template-operator/blob/main/tox.ini) suggests otherwise. Do you know how others think about this? And would there be push back to change it in operator template?

.github/workflows/integration.yaml
ca-scribner: aha!  [this](https://github.com/canonical/istio-operators/blob/a69721f08e9f627e3d29178a5f42dfcbf3405ee9/.github/workflows/integration.yaml#L43) is what I was thinking of.  No need to `run`, there's an arg instead

charms/mlflow-server/tox.ini
ca-scribner: I thought you might have gotten it from there.  Yeah I suggest you open a PR on the template about this.  Actually I could have sworn I already had, but I don't see one.  The positives feel way more fit for purpose than the negatives - I can't think of when the negatives would be better

.github/workflows/integration.yaml
DomFleischmann: I don't think this argument is available in `actions-operator` the action it is using in your example is `balchua/microk8s-actions@v0.2.2`.

ca-scribner: oh whoops yeah that's my fault.  Seems like something we should get into `actions-operator` then rather than `run` `microk8s` commands here

charms/mlflow-server/tests/integration/test_charm.py
```diff
@@ -0,0 +1,132 @@
...
+        namespace=ops_test.model_name,
+    )
+
+    await ops_test.model.set_config({"update-status-hook-interval": "15s"})
+    istio_gateway_role_name = "istio-gateway-operator"
+
+    new_policy_rule = PolicyRule(verbs=["*"], apiGroups=["*"], resources=["*"])
+    this_role = lightkube_client.get(Role, istio_gateway_role_name)
+    this_role.rules.append(new_policy_rule)
+    lightkube_client.patch(Role, istio_gateway_role_name, this_role)
```
ca-scribner, line 74: Can we add a comment here saying why we do this and under what condition we can stop?  It is essentially covering for a bug in istio-gateway, so we should say that.

Maybe something like "# Patch the istio-gateway Role so that it can access it's own configmap.  This can be removed when we move to the sidecar istio v1.11 charm" or similar

ca-scribner (approved):

PR #34, opened 2022-03-15
feat: check and auto-create S3 artifact bucket if missing

### LINKED ISSUE #23: MLFlow does not create a bucket when relation with object store is created.
When MLFlow is related to the object storage no default bucket is created. The bucket should be called "mlflow".

The result of missing bucket occurs on `mlflow.sklearn.log_model`:
```
S3UploadFailedError: Failed to upload /tmp/tmp16ktbcdm/model/MLmodel to mlflow/0/89195447e0d545e2a5d15e183d84e53a/artifacts/model/MLmodel: An error occurred (NoSuchBucket) when calling the PutObject operation: The specified bucket does not exist
```

**Workaround**
```
pip install awscli -q
aws --endpoint-url $MLFLOW_S3_ENDPOINT_URL s3 mb s3://mlflow
```

ca-scribner:
Adds the feature that:
* Checks related S3 storage for the bucket named by the default_artifact_root config option
* If create_artifact_root_if_not_exists==True and default bucket does not exist, attempts to create the bucket

Supporting this is an S3 bucket wrapper to help checking the bucket existance/creating new buckets and unit/integration tests.

Also included here was some refactoring to make the main function of the Operator read a bit clearer, packaging logic into some verbosely named helpers.

 Closes #23


ca-scribner:
# Test instructions:

```
# Build local charm, then
juju add-model mlflow # Needs to be named like this for the existing selenium tests
juju deploy ./mlflow-server_ubuntu-20.04-amd64.charm --resource oci-image=$(yq eval '.resources.oci-image.upstream-source' ./metadata.yaml)

juju deploy cs:~charmed-osm/mariadb-k8s-35 mysql-mlflow
juju deploy minio minio-mlflow --config access-key=minio --config secret-key=minio

# Once minio is fully deployed, demonstrate that minio has no buckets:
juju wait -wv

export MINIO_IP=$(juju show-unit minio-mlflow/0 | yq e '.minio-mlflow/0.address')
export MINIO_PORT=$(juju config minio-mlflow port)    
export MINIO_ACCESS_KEY=$(juju config minio-mlflow access-key) 
export MINIO_SECRET_KEY=$(juju config minio-mlflow secret-key)
kubectl run minio-client -i --tty --rm --image minio/mc -n mlflow --command -- /bin/sh -c "\
  mc alias set myMinio http://$MINIO_IP:$MINIO_PORT $MINIO_ACCESS_KEY $MINIO_SECRET_KEY
  echo Buckets in Minio:
  mc ls myMinio
  sleep 2  # k8s got cranky if this exited too fast
"
echo Output from pod:
kubectl logs minio-client

# Above should have no buckets in the output

# Relate mlflow to is requirements, and it'll fire up and create the "mlflow" bucket
juju relate mlflow-server mysql-mlflow
juju relate mlflow-server:object-storage minio-mlflow:object-storage

# After started, demonstrate the `mlflow` bucket now exists in minio

kubectl run minio-client -i --tty --rm --image minio/mc -n mlflow --command -- /bin/sh -c "\
  mc alias set myMinio http://$MINIO_IP:$MINIO_PORT $MINIO_ACCESS_KEY $MINIO_SECRET_KEY

  echo Buckets in Minio:
  mc ls myMinio

  sleep 2  # k8s got cranky if this exited too fast
"

echo Output from pod:
kubectl logs minio-client

# Above should show a bucket called mlflow
```

charms/mlflow-server/config.yaml
```diff
@@ -2,6 +2,12 @@
 # See LICENSE file for licensing details.
 #
 options:
+  create_default_artifact_root_if_missing:
+    description: |
+      If True, charm will try to create the default_artifact_root bucket in S3 if it does not
```
agathanatasha, line 7: is there a scenario where this is desirable?

ca-scribner: I'm not sure.  The case I thought about was if we're in gateway mode (eg: Minio is just a passthrough to storage on say AWS), I could see someone wanting to manage the bucket in the host rather than have this charm try to create things.  But that's also a bit of a stretch so I don't know if it is really needed or not.  What do you think?

ca-scribner: Yeah sorry I should have said if minio was in gateway mode and used as a passthrough to another blob store.  And yes, we do support/use it (or at least the next customer will).  And no, I did not know it was being deprecated :fearful:

DomFleischmann: Looking at the blocked status message which is very detailed I think this config option is okay as it is.

agathanatasha: I see. Since we support it, we should keep this config

agathanatasha (approved):

PR #49, opened 2022-08-15
feat: Add generic, clickable publish.yaml

ca-scribner:
Adds a generic publish.yaml that can be triggered through the UI

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,35 +1,82 @@
-# reusable workflow triggered by other actions
+# reusable workflow for publishing all charms in this repo
 name: Publish
 
 on:
   workflow_call:
     secrets:
-      charmcraft-credentials:
+      CHARMCRAFT_CREDENTIALS:
         required: true
+  workflow_dispatch:
+    inputs:
+      # TODO: Implement this
+      # source_github_branch:
+      #   description: Source branch to publish
+      #   required: false
+      #   default: 'main'
+      destination_channel:
+        description: CharmHub channel to publish to
+        required: false
+        default: 'latest/edge'
 
 jobs:
+  get-charm-paths:
+    name: Generate the Charm Matrix
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yaml
```diff
@@ -1,35 +1,82 @@
-# reusable workflow triggered by other actions
+# reusable workflow for publishing all charms in this repo
 name: Publish
 
 on:
   workflow_call:
     secrets:
-      charmcraft-credentials:
+      CHARMCRAFT_CREDENTIALS:
```
DnPlas, line 7: Is this to make a difference between integration and publish creds (which seem to be the same)?

DnPlas (approved):
Thanks @ca-scribner, I left just a question, other than that LGTM

.github/workflows/publish.yaml
ca-scribner: No, this is to reconcile a difference between how `workflow_call` and `workflow_dispatch` handle secrets:
* `workflow_call`: gets no access to this repo's secrets, but can accept [secrets as inputs](https://github.com/canonical/mlflow-operator/pull/49/files#diff-e81e13daadd1745de51d8b8d85fce1fcd9be355732b64d60a6b56e18c28388caR6)
* `workflow_dispatch`: gets full access to this repo's secrets, but **cannot** accept secrets as inputs

The nice thing is that any secrets passed to `workflow_call` are put into the same place as the regular github repository secrets would be (eg: you access them via `${{ secrets.SECRET_FROM_WORKFLOW_CALL }}`).  So if we name the `workflow_call` secret the same as the secret we have in our repo, the rest of the script can just call `${{ secrets.CHARMCRAFT_CREDENTIALS }}` and not know anything about the implementation

DnPlas: Oh shoot, I meant the difference between all caps `CHARMCRAFT_CREDENTIALS`

ca-scribner: @DnPlas and I discussed offline and we're now on the same page.  Resolving this conversation

PR #72, opened 2022-12-07
Pin to ubuntu-20.04 for CI

misohu:
Pin `runs-on` to `ubuntu-20.04` in every action. `latest` now points to `22.04`


i-chvets (approved):
Approved

i-chvets (approved):
Approved.

PR #92, opened 2023-01-12
fix: removed aws credentials from poddefault

i-chvets:
https://github.com/canonical/bundle-kubeflow/issues/528

Summary of changes:
- Removed AWS credentials from PodDefault.

### DIFF
--- a/charms/mlflow-server/src/charm.py
+++ b/charms/mlflow-server/src/charm.py
@@ -94,8 +94,6 @@ class Operator(CharmBase):
             {
                 "minio": {
                     "env": {
-                        "AWS_ACCESS_KEY_ID": obj_storage["access-key"],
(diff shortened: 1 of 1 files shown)


ca-scribner (approved):
I think you could remove the pod default entirely, but that leaving it does not break anything

i-chvets:
> Would this change mean that for MLFlow integration we need to also select "Allow Minio integration" too? Do we need to update the documentation for this?

I think, you are correct: we need to update documentation for this. I am not sure what you mean by "Allow Minio integration". Could you point me to doc/demo/tutorial for this?

i-chvets:
`mlflow-server-minio-credentials` secret needs to be present in user's namespace. We need to add this procedure to README.md

PR #94, opened 2023-01-16
fix: added unit test and updated comments

i-chvets:
Summary of changes:
- Added comments to S3 service code to clarify the usage.
- Added unit test to ensure that S3 URL endpoint is generated correctly in S3 service and in charm code.

### DIFF
--- a/charms/mlflow-server/src/charm.py
+++ b/charms/mlflow-server/src/charm.py
@@ -331,6 +331,7 @@ class Operator(CharmBase):
             {"name": f"{self.charm_name}-db-secret", "data": _db_secret_dict(mysql=mysql)},
         ]
 
+   @property
     def _gen_obj_storage_endpoint_url(self, obj_storage):
         """Generate object storage endpoint URL."""
         return f"http://{obj_storage['service']}.{obj_storage['namespace']}:{obj_storage['port']}"
--- a/charms/mlflow-server/src/charm.py
+++ b/charms/mlflow-server/src/charm.py
@@ -331,10 +331,20 @@ class Operator(CharmBase):
             {"name": f"{self.charm_name}-db-secret", "data": _db_secret_dict(mysql=mysql)},
         ]
 
-   @property
     def _gen_obj_storage_endpoint_url(self, obj_storage):
-        """Generate object storage endpoint URL."""
-        return f"http://{obj_storage['service']}.{obj_storage['namespace']}:{obj_storage['port']}"
+        """Generate object storage endpoint URL.
+
+        URL generated only if 'service' is set, otherwise it returns empty string.
+        """
+        endpoint_url = ""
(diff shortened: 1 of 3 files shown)


charms/mlflow-server/src/charm.py
```diff
@@ -331,6 +331,10 @@ def _define_secrets(self, obj_storage, mysql):
             {"name": f"{self.charm_name}-db-secret", "data": _db_secret_dict(mysql=mysql)},
         ]
 
+    def _gen_obj_storage_endpoint_url(self, obj_storage):
+        """Generate object storage endpoint URL."""
+        return f"http://{obj_storage['service']}.{obj_storage['namespace']}:{obj_storage['port']}"
```
DnPlas, line 336: Should we add a check to ensure  `obj_storage`  actually has those keys you are calling?

charms/mlflow-server/src/charm.py
```diff
@@ -331,6 +331,10 @@ def _define_secrets(self, obj_storage, mysql):
             {"name": f"{self.charm_name}-db-secret", "data": _db_secret_dict(mysql=mysql)},
         ]
 
+    def _gen_obj_storage_endpoint_url(self, obj_storage):
```
DnPlas, line 334:
```suggestion
   @property
    def _gen_obj_storage_endpoint_url(self, obj_storage):
```

charms/mlflow-server/tests/unit/test_operator.py
```diff
@@ -464,3 +465,18 @@ def test_install_without_nodeport_and_lb_services(harness, mocker):
     assert harness.charm.model.unit.status == ActiveStatus()
 
     assert pod_spec[0]["kubernetesResources"]["services"] == []
+
+
+def test_storage_endpoint_generation(harness, sample_object_storage):
+    """Test S3 storage URL for correct generation in S3 service and in charm code."""
+    harness.begin()
+    s3_wrapper = S3BucketWrapper(
```
DnPlas, line 473: Is there a way to mock this out?

DnPlas (commented):
Thanks @i-chvets, left some comments.

DnPlas (commented):
Thanks @i-chvets, left some comments.

charms/mlflow-server/src/charm.py
i-chvets: Older code never had those checks. But I guess it is good idea. Will return empty string then.

charms/mlflow-server/src/charm.py
i-chvets: This is not really a property. It is a generating function. So, I would not say this should be here.

charms/mlflow-server/src/charm.py
i-chvets: Done.

DnPlas (approved):

PR #95, opened 2023-01-20
Mlflow 2.1.1 custom image

misohu:


charms/mlflow-server/metadata.yaml
```diff
@@ -11,7 +11,7 @@ resources:
   oci-image:
     type: oci-image
     description: Backing OCI image
-    upstream-source: quay.io/helix-ml/mlflow:1.13.1
+    upstream-source: docker.io/misohu/mlflow-tracking-server:2.1.1_test
```
i-chvets, line 14: Can we find a proper place to put this image?
Do we have one for Charmed Kufeflow?
@DomFleischmann

PR #103, opened 2023-02-23
Run CI in AWS EC2 instance

misohu:
1. **Change CI to execute in AWS** - there is no space in the github runner to run the newly created mlflow image
2. **Move the charm code to root folder** - MLflow 2.1.1. will be a standalone charm no point to store code in charms subfolder
3. **Remove examples folder** - as we use 2.1.1. examples will change

The CI uses [this action](https://github.com/machulav/ec2-github-runner) which will create a ec2 instance which was created for the AWS service account (kubeflow-team). All the AWS credentials and GH PAT are part of the secrets (go to settings -> secrets and variables -> actions). All the AWS EC2 configuration is part of the repo variables (go to settings -> secrets and variables -> actions).

CI uses custom AMI created from ubuntu LTS 20.04 AMI and installing docker inside. The AMI is part of the service account private AMI list (its the value set in environment variables).


.github/workflows/integrate.yaml
```diff
@@ -4,92 +4,121 @@ name: CI
...
-        microk8s-addons: "dns storage rbac metallb:10.64.140.43-10.64.140.49"
-    - name: Install firefox
-      run: sudo apt install -y firefox
+    - name: Check out repo
+      uses: actions/checkout@v2
 
+    - name: Setup operator environment
+      # The runner will run the code as root. There is problem when using juju under rood so here is small workaround to exec as ubuntu user
+      run: |
+        chmod 777 .
```
DnPlas, line 79: Can we limit permissions or does it strictly have to be `777`?

.github/workflows/integrate.yaml
```diff
@@ -4,92 +4,121 @@ name: CI
...
-        # Pinned until this bug is resolved: https://bugs.launchpad.net/juju/+bug/1992833
-        bootstrap-options: "--agent-version=2.9.34"
-        microk8s-addons: "dns storage rbac metallb:10.64.140.43-10.64.140.49"
-    - name: Install firefox
-      run: sudo apt install -y firefox
+    - name: Check out repo
+      uses: actions/checkout@v2
 
+    - name: Setup operator environment
+      # The runner will run the code as root. There is problem when using juju under rood so here is small workaround to exec as ubuntu user
```
DnPlas, line 77:
```suggestion
      # The runner will run the code as root. There is problem when using juju under root so here is small workaround to exec as ubuntu user
```

.github/workflows/publish.yaml
```diff
@@ -26,37 +36,61 @@ on:
...
-
     steps:
       - name: Checkout
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
           ref: ${{ inputs.source_branch }}
-
+      - name: Setup environment
+        run: apt install -y python3 python-dev python3-dev build-essential libssl-dev libffi-dev libxml2-dev libxslt1-dev zlib1g-dev python3-pip
```
DnPlas, line 93: Does this work without `sudo`?

.github/workflows/setup_environment.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+# Install packages
+sudo snap install core
+sudo apt-get remove -qy lxd lxd-client
+sudo snap install lxd --channel=latest/stable
+sudo snap refresh lxd --channel=latest/stable
```
DnPlas, line 7: Does this have any effect?

.github/workflows/setup_environment.sh
```diff
@@ -0,0 +1,35 @@
...
+sudo apt-get update -yqq
+sudo apt-get install -yqq python3-pip
+sudo --preserve-env=http_proxy,https_proxy,no_proxy pip3 install tox
+
+# Juju + charmcraft + debug tools installation
+sudo snap install juju --classic --channel=latest/stable
+sudo snap install jq
+sudo snap install charm --classic --channel=latest/stable
+sudo snap install charmcraft --classic --channel=latest/stable
+sudo snap install juju-bundle --classic --channel=latest/stable
```
DnPlas, line 27: I do not think you are using `juju-bundle` at all.

.github/workflows/setup_environment.sh
```diff
@@ -0,0 +1,35 @@
...
+sudo snap install juju-crashdump --classic --channel=latest/stable
+
+# Microk8s setup
+sudo snap install microk8s --channel=1.22/stable --classic
```
DnPlas, line 31: We are testing 1.7 with microk8s 1.24, will MlFlow work on that version? If so, let's change this one.

```suggestion
sudo snap install microk8s --channel=1.24/stable --classic
```

.github/workflows/setup_environment.sh
```diff
@@ -0,0 +1,35 @@
...
+sudo snap refresh charmcraft --channel latest/candidate
+sudo usermod -a -G microk8s ubuntu
+sudo chown -f -R ubuntu /home/ubuntu/.kube
```
DnPlas, line 34: You are `chown`ing a dir that does not exist.

tox.ini
```diff
@@ -1,18 +1,80 @@
+# Copyright 2022 Canonical Ltd.
```
DnPlas, line 1:
```suggestion
# Copyright 2023 Canonical Ltd.
```

DnPlas (commented):
Thanks @misohu, few comments.

.github/workflows/integrate.yaml
```diff
@@ -4,92 +4,121 @@ name: CI
...
-      uses: actions/upload-artifact@v2
-      with:
-        name: selenium-har
-        path: /tmp/selenium-*.har
+    - name: Get juju status
+      run: /bin/su -c "juju status" - ubuntu 
       if: failure()
+
+    - name: Get workload logs
+      run: /bin/su -c "kubectl logs --tail 100 -ntesting -lapp.kubernetes.io/name=seldon-controller-manager" - ubuntu
```
DnPlas, line 96: Why seldon?

ca-scribner:
Don't refactor the current PR, but for a future thought my first impression here is I see both CI changes (aws stuff) and repo structure changes (moving charm from `./charms/mlflowserver` to `./`).  Those sound like good things to separate into different PRs - of 43 files changed, about 30 of them are file moves.  (I might have missed a nuance for this case, but ... ) if the repo structure change is a separate PR that only moves the files plus patches the old CI to work with the new file locations (only minimal changes to keep the old functionality), that's a really easy review because we can see the old functionality is maintained.  As is, we have to look at those moves and new CI and decide jointly if everything is ok

.github/workflows/integrate.yaml
```diff
@@ -4,92 +4,121 @@ name: CI
...
-        fetch-depth: 0
-    - name: Check libs
-      uses: canonical/charming-actions/check-libraries@2.1.1
-      with:
-        charm-path: ./charms/${{ matrix.charm }}
-        credentials: "${{ secrets.charmcraft-credentials }}"
-        github-token: "${{ secrets.GITHUB_TOKEN }}"
-
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
```
ca-scribner, line 29: Let's fully pin, and then renovate will manage this

```suggestion
        uses: aws-actions/configure-aws-credentials@v1.7.0
```

.github/workflows/integrate.yaml
```diff
@@ -4,92 +4,121 @@ name: CI
...
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+      - name: Start EC2 runner
+        id: start-ec2-runner
+        uses: machulav/ec2-github-runner@v2
```
ca-scribner, line 36:
```suggestion
        uses: machulav/ec2-github-runner@v2.3.3
```

.github/workflows/integrate.yaml
```diff
@@ -4,92 +4,121 @@ name: CI
...
-    name: Lint
-    runs-on: ubuntu-20.04
-    strategy:
-      fail-fast: false
-      matrix:
-        charm:
-        - server
+    name: Run lint
+    needs: start-runner
+    runs-on: ${{ needs.start-runner.outputs.label }}
```
ca-scribner, line 47: Does this need to run on the aws runner?  I think it can use a standard runner and remove the dependency on aws

.github/workflows/integrate.yaml
```diff
@@ -4,92 +4,121 @@ name: CI
...
+    - name: Check out repo
+      uses: actions/checkout@v2
+    - name: Install dependencies
+      run: sudo apt-get install -y python3-pip tox
+    - name: Lint code
+      run: tox -e lint
+  unit-tests:
+    name: Run unit tests
+    needs: [start-runner, lint]  
+    runs-on: ${{ needs.start-runner.outputs.label }}
```
ca-scribner, line 58: Does this need to run on the aws runner?  I think it can use a standard runner and remove the dependency on aws

.github/workflows/integrate.yaml
```diff
@@ -4,92 +4,121 @@ name: CI
...
+    - name: Get juju status
+      run: /bin/su -c "juju status" - ubuntu 
       if: failure()
+
+    - name: Get workload logs
+      run: /bin/su -c "kubectl logs --tail 100 -ntesting -lapp.kubernetes.io/name=seldon-controller-manager" - ubuntu 
+      if: failure()
+
+    - name: Get operator logs
+      run: /bin/su -c "kubectl logs --tail 100 -ntesting -loperator.juju.is/name=seldon-controller-manager" - ubuntu
```
ca-scribner, line 100: same as above.  copy/paste error?

.github/workflows/on_pull_request.yaml
```diff
@@ -13,11 +13,22 @@ jobs:
...
-      charmcraft-credentials: ${{ secrets.CHARMCRAFT_CREDENTIALS }}
+      CHARMCRAFT_CREDENTIALS: ${{ secrets.CHARMCRAFT_CREDENTIALS }}
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+      AWS_REGION: ${{ secrets.AWS_REGION }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      GH_PERSONAL_ACCESS_TOKEN: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
 
   # publish runs in parallel with tests, as we always publish in this situation
+  # For AWS the upload-charm command fails. Commenting for now
```
ca-scribner, line 24: Is this comment still relevant?  I don't see what is commented

.github/workflows/publish.yaml
```diff
@@ -26,37 +36,61 @@ on:
...
+          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+          ec2-image-id: ${{ vars.AWS_EC2_IMAGE_ID }}
+          ec2-instance-type: ${{ vars.AWS_EC2_INSTANCE_TYPE }}
+          subnet-id: ${{ vars.AWS_DEFAULT_SUBNET_ID }}
+          security-group-id: ${{ vars.AWS_SECURITY_GROUP_ID }}
   get-charm-paths:
     name: Generate the Charm Matrix
-    runs-on: ubuntu-20.04
+    needs: start-runner
+    runs-on: ${{ needs.start-runner.outputs.label }}
```
ca-scribner, line 65: Does this need to be on the custom runner and depend on `start-runner`?

.github/workflows/setup_environment.sh
```diff
@@ -0,0 +1,35 @@
...
+lxd waitready
+lxd init --auto
+chmod a+wr /var/snap/lxd/common/lxd/unix.socket
+lxc network set lxdbr0 ipv6.address none
+sudo usermod -a -G lxd ubuntu
+newgrp lxd
+
+# Python setup
+sudo apt-get update -yqq
+sudo apt-get install -yqq python3-pip
```
ca-scribner, line 19: To be consistent for other charms, do we need to explicitly call out python 3.8 here?  Or does this guarantee a python 3.8 install?  If this will always give python 3.8, maybe still comment on that here since we're particular about the version due to pip-compile+pylibjuju?

.github/workflows/setup_microk8s.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# This needs to be a separate file as newgrp will not have effect from  setup_environment.sh
+microk8s enable rbac dns storage metallb:10.64.140.43-10.64.140.49
+microk8s kubectl rollout status deployment/hostpath-provisioner -n kube-system
+
+# Juju controller + model
+juju bootstrap --agent-version=2.9.34 --no-gui microk8s uk8sx
+juju add-model kubeflow
```
ca-scribner, line 9: If accepting this, also need to change it in the integration tests where they use the kubeflow model for execution and dumping logs

```suggestion
juju add-model mlflow
```

lib/charms/observability_libs/v1/kubernetes_service_patch.py
```diff
@@ -0,0 +1,341 @@
+# Copyright 2021 Canonical Ltd.
```
ca-scribner, line 1: I don't think this gets used in this PR, does it?  It looks like it is being added, but charm.py is not changing

ca-scribner (changes requested):
Some changes.  Also, should integration tests be failing during charm build?

ca-scribner (commented):
(deleted)

.github/workflows/integrate.yaml
misohu: no User is also creating .tox folder so it needs to be 777 (I tried but not enough otherwise)

.github/workflows/publish.yaml
misohu: Runner in this case runs as root thats why the previous run as Ubuntu ... in this case is root

.github/workflows/setup_environment.sh
misohu: I saw it in operator action .... not sure I believe this is preventing some weird snap install behavior

.github/workflows/setup_microk8s.sh
misohu: Dont know about this one ... we potentially want to run it with kubeflow in same model.

lib/charms/observability_libs/v1/kubernetes_service_patch.py
misohu: Prometheus lib uses the 

```
from charms.observability_libs.v0.juju_topology import JujuTopology
```

.github/workflows/setup_environment.sh
misohu: removed thanks :)

.github/workflows/setup_environment.sh
misohu: fixed thanks :)

.github/workflows/integrate.yaml
misohu: fixed

.github/workflows/publish.yaml
misohu: after speaking with @beliaev-maksim we agreed that we will leave this one on aws ... in next task I will create the separate tasks for aws and run this in github-runner solely.

.github/workflows/integrate.yaml
misohu: after speaking with @beliaev-maksim we agreed that we will leave this one on aws ... in next task I will create the separate tasks for aws and run this in github-runner solely.

.github/workflows/integrate.yaml
misohu: after speaking with @beliaev-maksim we agreed that we will leave this one on aws ... in next task I will create the separate tasks for aws and run this in github-runner solely.

PR #104, opened 2023-02-23
Standalone sidecar charm

misohu:
Deploy MLflow as a sidecar charm (without any relation).

MLflow is running and is exposed as NodePort on port speciffied in config. 

This is a follow up PR on https://github.com/canonical/mlflow-operator/pull/103 (first review the 103).


metadata.yaml
```diff
@@ -12,7 +12,7 @@ resources:
   oci-image:
     type: oci-image
     description: Backing OCI image
-    upstream-source: docker.io/charmed/base-mlflow
+    upstream-source: docker.io/charmedkubeflow/mlflow:latest
```
DnPlas, line 15: Should we use a tag here? Maybe we can keep a 1:1 relation with upstream in terms of image tags, in the end this image corresponds to a certain mlflow version, no?

DnPlas (commented):
Thansk @misohu, I reviewed commits 9d03e6228fd2bc20f1b2b4d4f857d0983ebbe190 and e9ee80ee2b4dff06b696539cf24b17a941514669, just left a comment.

metadata.yaml
misohu: for now we have just latest as we need to adjust the image's CI .... we cant use upstream tag as we may want to have multiple versions of upstream tag in case we will fix vulnerabilities. There is a task in jira for that for now we have just latest tag.

src/charm.py
```diff
@@ -1,427 +1,127 @@
...
+from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
+from ops.pebble import ChangeError, Layer
 
-class Operator(CharmBase):
-    """Charm for the ML Flow Server.
 
-    https://github.com/canonical/mlflow-operator
-    """
+class MlflowCharm(CharmBase):
+    """A Juju Charm for Training Operator"""
```
ca-scribner, line 18:
```suggestion
    """A Juju Charm for MLFlow."""
```

ca-scribner: Can you check your linting config vs our other charms?  I'd have expected this to fail because it doesn't end in ".", so I think the config is not the same as our other charms

src/charm.py
```diff
@@ -1,427 +1,127 @@
...
-                            {
-                                "protocol": "TCP",
-                                "port": config["mlflow_port"],
-                                "targetPort": config["mlflow_port"],
-                                "nodePort": config["mlflow_nodeport"],
-                            }
-                        ],
-                    },
-                }
+            self.logger.info("I am Michal Hucko")
```
ca-scribner, line 41: Good to know, but probably not needed :)

```suggestion
```

src/charm.py
```diff
@@ -1,427 +1,127 @@
...
+    @property
+    def container(self):
+        """Return container."""
+        return self._container
 
-    def _get_pod_spec_services(self, config):
-        """Returns service list for pod spec based on enabled service flags."""
-        pod_spec_services = []
+    def _create_service(self):
         if self.config["enable_mlflow_nodeport"]:
```
ca-scribner, line 40: This if/else might be more readable if it only set the stuff that differs between the two branches, rather than have it all duplicated.  I think atm it is somehing like:
```python
if True:
  x = "true branch"
  y = "true branch"
  do_something(
    common_setting,
    x,
    y
  )
else:
  x = "false branch"
  y = "false branch"
  do_something(
    common_setting,
    x,
    y
  )
```
which makes the reader need to focus to understand what is different.  What about writing it like:
```python
if True:
  x = "true branch"
  y = "true branch"
else:
  x = "false branch"
  y = "false branch"
do_something(
  common_setting,
  x,
  y
)
```
that way the common parts are easy to spot

src/charm.py
```diff
@@ -1,427 +1,127 @@
...
+                self.container.replan()
+            except ChangeError as err:
+                raise ErrorWithStatus(f"Failed to replan with error: {str(err)}", BlockedStatus)
+
+    def _on_pebble_ready(self, _):
+        """Configure started container."""
+        if not self.container.can_connect():
+            # Pebble Ready event should indicate that container is available
+            raise ErrorWithStatus("Pebble is ready and container is not ready", BlockedStatus)
+
```
ca-scribner, line 110: Should this be an Error rather than just something Blocked?  Blocked suggests it is something that a user can intervene about (change a setting, etc), but I don't think that's the case here

src/charm.py
```diff
@@ -1,427 +1,127 @@
...
+        current_layer = self.container.get_plan()
+        new_layer = self._charmed_mlflow_layer()
+        if current_layer.services != new_layer.services:
+            self.unit.status = MaintenanceStatus("Applying new pebble layer")
+            self.container.add_layer(self._container_name, new_layer, combine=True)
+            try:
+                self.logger.info("Pebble plan updated with new configuration, replaning")
+                self.container.replan()
+            except ChangeError as err:
+                raise ErrorWithStatus(f"Failed to replan with error: {str(err)}", BlockedStatus)
```
ca-scribner, line 103: Should this be an Error rather than just something Blocked? Blocked suggests it is something that a user can intervene about (change a setting, etc), but I don't think that's the case here.

src/charm.py
```diff
@@ -1,427 +1,127 @@
...
-            f"as expected. Caught exception: '{str(e)}'",
-            BlockedStatus,
-        )
-
-    return obj_storage
+            self._check_leader()
+            self._update_layer()
+        except ErrorWithStatus as err:
+            self.model.unit.status = err.status
+            self.logger.info(f"Failed to handle {event} with error: {str(err)}")
```
ca-scribner, line 121: I think this log message is too aggressive.  We're not failing to handle errors here, we get here by *handling* errors.  For example, if this unit is not the leader it is going to say "Failed to handle X with error: 'Not a leader, skipping setup'" and go into `WaitingStatus`.  I like that we log the message, but not that we call it an unhandled error
```suggestion
            self.logger.info(f"Event {event} stopped early with message: {str(err)}")
```

tests/unit/test_operator.py
```diff
@@ -1,482 +1,80 @@
...
-    assert s3_wrapper.s3_url == harness.charm._gen_obj_storage_endpoint_url(
-        obj_storage=sample_object_storage
+    @patch(
+        "charm.KubernetesServicePatch",
+        lambda x, y, service_name, service_type, refresh_event: None,
     )
+    def test_update_layer_success(
+        self,
+        harness: Harness,
+    ):
```
ca-scribner, line 77: Is there a better name we can give this?  It tests more than just "update_layer_success"

ca-scribner (commented):
minor stuff, otherwise lgtm

src/charm.py
misohu: Blocked is consistent with rest of our rewrites .... do we want to change it in every sidecar charm? If so Lets keep this Blocked and create a task. What do you think?

src/charm.py
misohu: Blocked is consistent with rest of our rewrites .... do we want to change it in every sidecar charm? If so Lets keep this Blocked and create a task. What do you think?

src/charm.py
misohu: Solved thanks :)

src/charm.py
DnPlas: I agree, a `BlockedStatus` means the charm needs a human operator to unblock, a `WaitingStatus` could be more appropriate here. I think this should be fixed before we merge your changes from the dev branch into main, but this PR is already very complex and convoluted. I am okay if you send a follow up PR just for this.

src/charm.py
DnPlas: I agree, this should get the unit in `Error` state, let's work on it in a separate PR before merging all the changes from dev branch into main.

DnPlas (approved):
Thanks @misohu, I think there are a few comments that we can resolve in a separate PR as this one is already very complex and convoluted. Please make sure to resolve the conversations of the status before we merge your dev branch into main.

PR #105, opened 2023-02-24
Sidecar charm with relations v211

misohu:
Add relation to minio to the charm
Add relation to mysql to the charm
Create minio bucket if not created

This is a follow up PR on https://github.com/canonical/mlflow-operator/pull/103 and https://github.com/canonical/mlflow-operator/pull/104 review this one after merging 103 and 104.


src/charm.py
```diff
@@ -74,25 +79,133 @@ def _charmed_mlflow_layer(self) -> Layer:
...
+                        f"cannot be created.  Caught error: '{str(e)}",
+                        BlockedStatus,
+                    )
+            else:
+                raise ErrorWithStatus(
+                    "Error with default S3 artifact store - bucket not accessible or does not "
+                    "exist. Set create_default_artifact_root_if_missing=True to automatically "
+                    "create a missing default bucket",
+                    BlockedStatus,
+                )
```
DnPlas, line 197: I'd split this method into two:

One for validating the name and access:

```python
def _validate_default_s3_bucket_name_and_access(self, bucket_name: str, s3_wrapper: S3BucketWrapper) -> bool:
  """Validates the default s3 bucket name is valid and the bucket is accessible.
        If it is not accessible and the `create_default_artifact_root_if_missing` config value
        is True, returns False; True otherwise.

         Args:
           bucket_name: ...
           s3_wrapper: ...
         Raises:
           ErrorWithStatus ...
             
  """
  ...
  if not validate_s3_bucket_name(self.config["default_artifact_root"]):
     raise ErrorWithStatus(...)

  if not s3_wrapper.check_if_bucket_accessible(...) and not self.config["create_default_artifact_root_if_missing"]:
    raise ErrorWithStatus("Bucket not accessible or does not exist...")
  elif not s3_wrapper.check_if_bucket_accessible(...) and self.config["create_default_artifact_root_if_missing"]:
    return False
  return True
```

Then have a dedicated method for creating the bucket:

```python
def _create_default_s3_bucket(self, s3_wrapper: S3BucketWrapper) -> None:
  """Creates an s3 bucket using the default_artifact_root config value.
        Raises:
          ErrorWithStatus: ...
  """
  try:
    s3_wrapper.create_bucket(self.config["default_artifact_root"])
  except:
    raise ErrorWithStatus(...)
```

Then you could use them in `_on_event()`:

```python
s3_wrapper = S3BucketWrapper(
            access_key=obj_storage_data["access-key"],
            secret_access_key=obj_storage_data["secret-key"],
            s3_service=f"{obj_storage_data['service']}.{obj_storage_data['namespace']}",
            s3_port=obj_storage_data["port"],
        )
if not self.validate_s3_bucket_name(bucket_name=self.config["default_artifact_root"], s3_wrapper=s3_wrapper):
   self._create_s3_default_buckets3_wrapper
default_artifact_root = self.config["default_artifact_rool"] # I believe we don't change this anywhere else, the name will always be this (?)
```

src/services/s3.py
```diff
@@ -0,0 +1,91 @@
+"""Wrapper for basic accessing and validating of S3 Buckets."""
```
DnPlas, line 1: Nice addition

DnPlas (commented):
Thanks @misohu, I just reviewed the charm code and left a comment, will check the test files later.

src/charm.py
misohu: Added your comment and fixed the tests :) thanks

src/charm.py
```diff
@@ -1,427 +1,269 @@
...
-            },
-            {"name": f"{self.charm_name}-db-secret", "data": _db_secret_dict(mysql=mysql)},
-        ]
+            relational_db = relational_db[0]
+            db_unit = list(relational_db.units)[0]
+            relational_db = relational_db.data[db_unit]
+            relational_db["database"]
+            return relational_db
+        except (IndexError, KeyError):
+            raise ErrorWithStatus("Waiting for relational-db relation data", WaitingStatus)
```
ca-scribner, line 125: What about splitting these up to be more descriptive?  I think this catches both:
* we have no related charm on `relational-db`
* we have a related charm, but it hasn't sent us the data we expected

Can we parse these separately to be clearer about this?  It feels like it would be hard to debug.  I think `IndexError` will always mean we have no relation, and `KeyError` that we have a relation but it hasn't sent us what we expect?

src/charm.py
```diff
@@ -1,427 +1,269 @@
...
+            raise ErrorWithStatus("Waiting for object-storage relation data", WaitingStatus)
 
+        try:
+            obj_storage = list(obj_storage.get_data().values())[0]
+        except Exception as e:
+            raise ErrorWithStatus(
+                f"Unexpected error unpacking object storage data - data format not "
+                f"as expected. Caught exception: '{str(e)}'",
+                BlockedStatus,
+            )
```
ca-scribner, line 142: This sounds like an Error rather than Blocked.  Unless it can happen in regular operation of the charm, and the charm can later resolve itself

src/charm.py
```diff
@@ -1,427 +1,269 @@
...
-    """Returns a dict of db-secret credential data, base64 encoded."""
-    db_secret = {
-        "DB_ROOT_PASSWORD": mysql["root_password"],
-        "MLFLOW_TRACKING_URI": f"mysql+pymysql://root:{mysql['root_password']}@{mysql['host']}"
-        f":{mysql['port']}/{mysql['database']}",
-    }
-    return _b64_encode_dict(db_secret)
+        return obj_storage
+
+    def _get_env_vars(self, relational_db_data, object_storage_data):
```
ca-scribner, line 146: This method is static - it doesn't need `self` so can either be a `@staticmethod` or a helper

ca-scribner (commented):
a few small things, otherwise lgtm

tests/integration/test_charm.py
```diff
@@ -152,152 +146,18 @@ async def does_minio_bucket_exist(bucket_name, ops_test: OpsTest):
...
-        else:
-            driver.get(url)
+async def test_can_create_experiment_with_mlflow_library(ops_test: OpsTest):
+    config = await ops_test.model.applications[CHARM_NAME].get_config()
+    url = f"http://localhost:{config['mlflow_nodeport']['value']}"
+    client = MlflowClient(tracking_uri=url)
+    response = requests.get(url)
+    client.create_experiment(TEST_EXPERIMENT_NAME)
+    all_experiments = client.search_experiments()
+    assert response.status_code == 200
```
ca-scribner, line 157: can we move this assertion up to the `response = ...` call?  If this gives `!=200`, the client calls don't matter

tests/integration/test_charm.py
```diff
@@ -152,152 +146,18 @@ async def does_minio_bucket_exist(bucket_name, ops_test: OpsTest):
...
-                    )
-                )
-                break
-            except WebDriverException:
-                sleep(5)
-        else:
-            driver.get(url)
+async def test_can_create_experiment_with_mlflow_library(ops_test: OpsTest):
+    config = await ops_test.model.applications[CHARM_NAME].get_config()
+    url = f"http://localhost:{config['mlflow_nodeport']['value']}"
```
ca-scribner, line 152: This test works for microk8s, but not for other k8s setups.  Is there a way this test can be executed in a k8s-agnostic way?

tests/unit/test_operator.py
```diff
@@ -1,482 +1,398 @@
...
-    assert (
-        b64decode(secrets_dict[f"{charm_name}-db-secret"]["data"]["DB_ROOT_PASSWORD"]).decode(
-            "utf-8"
+        object_storage_relation_id, "storage-provider", object_storage_data
+    )
+    return harness
+
+
+@pytest.fixture(scope="function")
+def relational_db_relation(obejct_storage_relation: Harness) -> Harness:
```
ca-scribner, line 82: This is really `relational_db_relation_with_object_storage_relation` - is there a better way to explain that up front to the reader?

tests/unit/test_operator.py
```diff
@@ -1,482 +1,398 @@
...
 
+    @patch(
+        "charm.KubernetesServicePatch",
+        lambda x, y, service_name, service_type, refresh_event: None,
+    )
+    @patch("charm.MlflowCharm._validate_default_s3_bucket_name_and_access")
+    def test_get_relational_db_data_success(
+        self, validate_default_s3_bucket: MagicMock, relational_db_relation: Harness
+    ):
+        relational_db_relation.begin_with_initial_hooks()
```
ca-scribner, line 227: Can we clarify the nomenclature here?  `relational_db_relation` is really a `harness` that has db and object storage relations, but that doesn't come across here.

src/charm.py
misohu: Blocked is consistent with rest of our rewrites .... do we want to change it in every sidecar charm? If so Lets keep this Blocked and create a task. What do you think?

tests/integration/test_charm.py
misohu: Also the bucket test uses microk8s ... do we want to do it in this PR ? There needs to be significant  change i guess :(

misohu: created issue: https://github.com/canonical/mlflow-operator/issues/106

src/charm.py
ca-scribner: Agreed

ca-scribner (approved):

tests/integration/test_charm.py
```diff
@@ -32,6 +44,120 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+    )
+    assert ops_test.model.applications[CHARM_NAME].units[0].workload_status == "waiting"
+    assert (
+        ops_test.model.applications[CHARM_NAME].units[0].workload_status_message
+        == "Waiting for object-storage relation data"
+    )
+
+
+@pytest.mark.abort_on_fail
+async def test_add_relational_db_with_relation_expect_active(ops_test: OpsTest):
```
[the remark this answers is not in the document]
ca-scribner: You're also welcome to merge this PR and then catch this one little fix separately as a small PR if that's easier for you.  This doesn't really block moving on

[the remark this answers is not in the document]
misohu: merging will fix it in this iteration task

PR #109, opened 2023-03-06
Run CI jobs in Runners and weekly checks in AWS

misohu:
Run CI in Github runners
Run scheduled weekly job with lates/edge in aws


tox.ini
```diff
@@ -74,7 +74,17 @@ deps =
...
+    pytest -v --tb native --asyncio-mode=auto {[vars]tst_path}integration/test_deploy_runners.py --keep-models --log-cli-level=INFO -s {posargs}
+    pytest -v --tb native --asyncio-mode=auto {[vars]tst_path}integration/test_charm.py --log-cli-level=INFO -s {posargs}
+deps =
+    -r requirements-integration.txt
+description = Run integration tests
+
+[testenv:aws-integration]
+commands =
+    pytest -v --tb native --asyncio-mode=auto {[vars]tst_path}integration/test_deploy_aws.py --keep-models --log-cli-level=INFO -s {posargs}
+    pytest -v --tb native --asyncio-mode=auto {[vars]tst_path}integration/test_charm.py --log-cli-level=INFO -s {posargs}
```
ca-scribner, line 87: In general, it is really helpful for maintenance if our `tox.ini` doesn't need special environments.  Ideally, we have a common `tox.ini` for all charms so we don't need to maintain multiple (ex: what obs does.  We've tried to do that too but haven't been rigid about it).

This feels like something that could be handled by an argument in pytest passed in through the `{posargs}`

ca-scribner: This could even still support multiple test files.  The arguments we pass here could be `-k` or `mark`'s.  

A vote in favor of editing the `tox.ini` like you have here though is that it makes it clearer to a user what is available.  Forcing users to pass `-k` and `mark`s makes them *know* they're available - that's not good either

tests/integration/test_deploy_aws.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class TestDeployAWS:
+    @pytest.mark.abort_on_fail
+    async def test_001_build_and_deploy(self, ops_test: OpsTest):
+        """Build and deploy the charm.
+
+        Assert on the unit status.
+        """
+        await ops_test.model.deploy(CHARM_NAME, channel="latest/edge", trust=True)
```
ca-scribner, line 18: Because there's one line of code different between `test_deploy_aws.py` and `test_charm.py`, this jumps out at me as a case where there's probably a better way.  Some ideas that come to mind:
* pass arguments into the test to change the behaviour.  Perhaps something like obs where they optionally pass the path to an already built charm?  Could do something like that with an optional path to a charm that needs pulling?  
* use marks or some other control flow built into pytest.  That at least avoids multiple test files.  You can have a default path for typical use, and a "please use aws instead" path for the special case
* either for marks or for separate files, you can still share the bulk of the test code.  Something like:

```python
def test_on_aws():
  some_configurations = aws_specific_configurations
  build_and_deploy_helper(some_configurations)

def test_normally():
  some_configurations = normal_configurations
  build_and_deploy_helper(some_configurations)

def build_and_deploy_helper(...):
  # this has all the ops_test.model.deploy, wait, etc stuff.  That way it is shared between both
```

beliaev-maksim: +1 to remove duplication via method extraction

however, not clear the idea of files. I think there is a value in keeping different files.
for example if you would like to create tests that run on AWS only, then having separate file for AWS will be cleaner

beliaev-maksim:
issue in the CI looks to be an issue in charming actions themselves, no?

tests/integration/test_deploy_aws.py
misohu: Yeah @ca-scribner  we had this discussion in bundle tests release tests where we decided marks are not a good way so we should rather use separate files .... I think separate files are good way here as there might be more testing connected with aws specifically.

tox.ini
misohu: Yeah @ca-scribner  we had this discussion in bundle tests release tests where we decided marks are not a good way so we should rather use separate files .... I think separate files are good way here as there might be more testing connected with aws specifically.

beliaev-maksim:
for history: CI fails due to merge into feature branch, see https://github.com/canonical/charming-actions/issues/88

tests/integration/test_deploy_aws.py
i-chvets: I am in for separate files. Makes it a bit more readable and independent. Also, if we grow test suites for AWS only or Azure only with specific tests, I wouldn't want to have them in regular integration tests. It will make regular integration test files for development too big and unreadable.

tox.ini
i-chvets: I think separate environments for AWS and regular integration are cleaner. Also, will allows do changes in tox.ini later on when we add some other test files or settings specific for AWS. Eg. run integration test on different K8S in AWS - CK8S and EKS - then that `integration-aws` environment will take care of the setup for each and will re-run the test.

ca-scribner: The benefit of consistent tox environments is that the github workflows for integration.yaml are then also common between different charms, meaning we don't need to maintain that file but instead share it across charms (and teams).  This was something @sed-i and I were discussing last week.  For the goal of reducing duplication between our charms and then also other team's charms, having the same tox environments is really helpful

That being said, this is a special case so maybe it makes sense to deviate here.

tests/integration/test_deploy_aws.py
ca-scribner: there's two orthogonal conversations here that can be decided separately:
* should we use helpers like demo'd above
* should we use marks or separate files

Sounds like we agree that we should use method extraction so that `test_on_aws()` and `test_normally()` differ only on the configuration they set, and then they share the helper for actually doing the building and deploying.  Agreed?

Re marks vs files: we go with separate files

tox.ini
ca-scribner: The benefit of consistent tox environments is that we can have common github workflow files (integration.yaml) across our charms (and across teams), reducing our duplication and maintenance.  This was something @sed-i raised recently when we were discussing how to find commonality in our tests.  For the goal of reducing duplication between our charms and then also other team's charms, having the same tox environments is really helpful

That being said, this is a special case so maybe it makes sense to deviate here.

tests/integration/test_deploy_aws.py
```diff
@@ -0,0 +1,26 @@
...
+
+import pytest
+import yaml
+from pytest_operator.plugin import OpsTest
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+CHARM_NAME = METADATA["name"]
+
+
+class TestDeployAWS:
```
sed-i, line 11: Haven't looked / just a fly-by comment:

Would you be able to keep only one tox env using a decorator such as
`@skipIf(os.environ.get("AWS_SOMETHING"), "Skipping because...")`

For example, we [can detect](https://github.com/canonical/traefik-k8s-operator/blob/a287fa6a41077c8c3b7d3eb244f055c2a354bd2a/tests/integration/conftest.py#L42) if we're running in a GH runner.

ca-scribner: I think @sed-i is on to something with this.  Feels like something that could be environment-aware

Although whether it is worth it, that's another question

tox.ini
i-chvets: Currently, we do not share `integration.yaml` between charms. They all different. There is no plan to unify it at this point. This will be a big effort and should be addressed as a separate roadmap item, speced out, and executed. At this present iteration I think we should proceed without trying to design for the future.

tests/integration/test_deploy_aws.py
i-chvets: I would disagree with this *test_on_aws() and test_normally() differ only on the configuration they set*.
What if my AWS tests will utilize AWS specific feature in the integration test, eg. use "awscli" or connect to AWS service?
Ideally, the interfaces we use should be abstract enough for us to not care about what cloud is out there, but reality is different as we know.

i-chvets: +1 for shaing helpers for building/deploying
+1 for anything that can be generalized to be move to `test_charm.py`

beliaev-maksim (approved):

misohu:
After talking to @i-chvets and @beliaev-maksim I will merge for now as it is, but we will write a spec on how we gonna approach testing in different clouds to fine tune.

PR #115, opened 2023-03-27
KF-739

misohu:
add secrets relation and send secrets through it


src/charm.py
```diff
@@ -236,6 +243,23 @@ def _on_pebble_ready(self, _):
...
+    def _send_manifests(self, interfaces, context, folder, relation):
+        """Send manifests from folder to desired relation."""
+        if relation in interfaces and interfaces[relation]:
+            manifests = self._create_manifests(folder, context)
+            interfaces[relation].send_data({relation: manifests})
+
+    def _create_manifests(self, folder, context):
+        """Create manifests string for given folder and context."""
+        manifests = []
+        manifest_files = glob.glob(f"{folder}/*.j2")
```
ca-scribner, line 255: I slightly favor writing these files out explicitly like [here](https://github.com/canonical/seldon-core-operator/blob/c34309f57403cd2a53d1216e816d52b083202b22/src/charm.py#L30) rather than slurping up the entire folder, mainly because I don't know if I'd expect adding a new file to that folder will immediately send it through this relation.  But if you prefer this way then lgtm

src/charm.py
```diff
@@ -257,6 +281,14 @@ def _on_event(self, event) -> None:
...
                 self._create_default_s3_bucket(s3_wrapper, bucket_name)
             self._update_layer(envs, bucket_name)
+            secrets_context = {
+                "s3_endpoint": f"http://{object_storage_data['service']}.{object_storage_data['namespace']}:{object_storage_data['port']}",  # noqa: E501
+                "s3_type": "s3",
+                "s3_provider": "minio",
+                "access_key": object_storage_data["access-key"],
+                "secret_access_key": object_storage_data["secret-key"],
+            }
+            self._send_manifests(interfaces, secrets_context, SECRETS_TEMPLATE_FOLDER, "secrets")
```
ca-scribner, line 291: This works here, but feels a bit heavy.  We render all manifests with the same context, so it relies on every .j2 file knowing how to avoid name collisions.  It is convenient to implement though.  If you can think of an easy way to avoid this it would be good, otherwise lgtm and we can just fix it if it becomes a problem.

ca-scribner (approved):
a few comments but you can ignore them if you choose.  Otherwise lgtm!

src/charm.py
beliaev-maksim: explicit better than implicit. I am +1 on names list

src/charm.py
misohu: Lets resolve this in poddefaults change as its more related to that :)

PR #123, opened 2023-04-14
KF-1031 Prometheus integration with alert rules

misohu:
- Add prometheus integration with basic alert rule and integraiton tests.


src/charm.py
```diff
@@ -49,6 +51,20 @@ def __init__(self, *args):
             self.framework.observe(self.on[rel].relation_changed, self._on_event)
         self._create_service()
 
+        # Prometheus related config
+        self.prometheus_provider = MetricsEndpointProvider(
+            charm=self,
+            relation_name="metrics-endpoint",
+            jobs=[
+                {
+                    "metrics_path": PROMETHEUS_PATH,
```
beliaev-maksim, line 60: do we really need it ?
it is the default value

src/charm.py
```diff
@@ -28,6 +29,7 @@
     "src/poddefaults/poddefault-minio.yaml.j2",
     "src/poddefaults/poddefault-mlflow.yaml.j2",
 ]
+PROMETHEUS_PATH = "/metrics"
```
beliaev-maksim, line 32: I would call is "SCRAPE_ENDPOINT"

beliaev-maksim: or prometheus scrape endpoint

beliaev-maksim (approved):
LGTM
maybe just the var name could be more explicit, but up to you

src/prometheus_alert_rules/unit_unavailable.rule
```diff
@@ -0,0 +1,10 @@
+alert: MLFlowServerUnitIsUnavailable
+expr: up < 1
+for: 0m
```
beliaev-maksim, line 3:
```suggestion
for: 5m
```

beliaev-maksim (approved):

PR #126, opened 2023-04-17
Kf 1003 mysql data platform library integration

misohu:


tests/integration/test_charm.py
```diff
@@ -143,9 +143,7 @@ async def test_add_relational_db_with_relation_expect_active(self, ops_test: Ops
             timeout=600,
         )
         await ops_test.model.relate(OBJECT_STORAGE_CHARM_NAME, CHARM_NAME)
-        await ops_test.model.relate(
-            f"{RELATIONAL_DB_CHARM_NAME}:mysql", f"{CHARM_NAME}:relational-db"
-        )
+        await ops_test.model.relate(f"{RELATIONAL_DB_CHARM_NAME}", f"{CHARM_NAME}")
```
beliaev-maksim, line 146:
```suggestion
        await ops_test.model.relate(RELATIONAL_DB_CHARM_NAME, CHARM_NAME)
```

beliaev-maksim (changes requested):
and another point, I do not see us monitoring the event if database endpoint changes

also, we need to put charm into Blocked state if DB relation is removed

src/charm.py
```diff
@@ -149,25 +159,22 @@ def _get_interfaces(self):
...
-            relational_db["database"]
-            return relational_db
-        except IndexError:
-            raise ErrorWithStatus("Waiting for relational-db relation data", WaitingStatus)
-        except KeyError:
-            raise ErrorWithStatus("Missing data in relational-db relation", WaitingStatus)
+    def _get_relational_db_data(self) -> dict:
+        data = self.database.fetch_relation_data()
+        self.logger.debug("Got following database data: %s", data)
+        for _, val in data.items():
```
beliaev-maksim, line 165:
```suggestion
        for val in data.values():
```

src/charm.py
```diff
@@ -149,25 +159,22 @@ def _get_interfaces(self):
...
+            self.logger.info("New mysql database endpoint is %s", val["endpoints"])
+            host, port = val["endpoints"].split(":")
+            db_data = {
+                "host": host,
+                "port": port,
+                "username": val["username"],
+                "password": val["password"],
+            }
+            return db_data
+        raise ErrorWithStatus("Waiting for relational-db relation data", WaitingStatus)
```
beliaev-maksim, line 177: what will happen if relation does not exist ?

can we identify that there is no relation and ask to add it?

misohu: well it will set status to waiting with `Waiting for relational-db relation data`. Do you want to change the text to sth else ?

src/charm.py
```diff
@@ -41,7 +42,11 @@ def __init__(self, *args):
         self.logger = logging.getLogger(__name__)
         self._port = self.model.config["mlflow_port"]
         self._container_name = "mlflow-server"
+        self._database_name = "mlflow"
```
i-chvets, line 45: Is it possible to make this configurable?

src/charm.py
```diff
@@ -92,8 +103,7 @@ def _create_service(self):
             refresh_event=self.on.config_changed,
         )
 
-    @staticmethod
-    def _get_env_vars(relational_db_data, object_storage_data):
+    def _get_env_vars(self, relational_db_data, object_storage_data):
         """Return environment variables based on model configuration."""
```
i-chvets, line 107: We started to covert this to property `service_environment`. Just a suggestion.

src/charm.py
i-chvets: Should we have Blocked status in this case with message `Please add required relational-db relation`?

src/charm.py
i-chvets, line 107: We started to covert this to property `service_environment`. Just a suggestion.

src/charm.py
i-chvets: I agree, we should add more logging/status setting:
- Relation does not exist: Blocked/Please add . . .
- Data is empty: Waiting/Found empty relation data  . . .
- Invalid data: Waiting/Found incorrect data  . . .

src/charm.py
```diff
@@ -9,6 +9,7 @@
 
 import yaml
 from charmed_kubeflow_chisme.exceptions import ErrorWithStatus
+from charms.data_platform_libs.v0.data_interfaces import DatabaseRequires
```
i-chvets, line 12: Is this database interface a required interface?

src/charm.py
misohu: Tbh I dont think there is usecase for different db name ... I am not sure ... Last time I was instructed not to overconfigure the charms :)

src/charm.py
misohu: Its used to create the database and register new events. So I think it is

src/charm.py
beliaev-maksim: agree with Michal.
Let's keep DB name solid

src/charm.py
misohu: To keep the pr small with just the context of the task I propose openning issue and facing it in another pr.

beliaev-maksim: @misohu 
you probably misunderstood

you just update method name and add `@property`

src/charm.py
beliaev-maksim: yes, required

src/charm.py
misohu: not that easy ... I need to create setter ... and fix 4 tests ... I will do it in PR for get_password action ok ? :)

beliaev-maksim: you do not need setter

beliaev-maksim: you will need to put relational_db_data, object_storage_data into the property itself, rather than pointing from outside

.github/workflows/integrate.yaml
```diff
@@ -60,6 +60,10 @@ jobs:
       # TODO: Unpin this when https://github.com/charmed-kubernetes/actions-operator/pull/46 is merged
       uses: claudiubelu/actions-operator@18ebf92ae3043bd3dd15238e5d9b662d7ba08daf
       with:
+        juju-channel:
+          description: "snap channel for juju, installed via snap"
+          required: false
+          default: "2.9/stable"
```
beliaev-maksim:
```suggestion
        juju-channel: "2.9/stable"
```

.github/workflows/integrate.yaml
```diff
@@ -60,6 +60,7 @@ jobs:
       # TODO: Unpin this when https://github.com/charmed-kubernetes/actions-operator/pull/46 is merged
       uses: claudiubelu/actions-operator@18ebf92ae3043bd3dd15238e5d9b662d7ba08daf
```
beliaev-maksim, line 61:
```suggestion
      uses: charmed-kubernetes/actions-operator@1.1.0
```

src/charm.py
beliaev-maksim: talked with Michal, too much refactoring for the current PR. Handle it in another PR

i-chvets: Sounds good.

src/charm.py
i-chvets: Sounds good. I will also hard code it in mine, to keep uniform.
The only usecase I can think of is customer having single MySQL installation with different DBs (different DB names) and related to different MLflow charms OR already having DB with name `mlflow` in MySQL DB.
There is an issue with hardcoded S3 bucket naming in some of our other charms So, same goes for DB names.

src/charm.py
i-chvets: Sounds goof.

src/charm.py
misohu: Added

src/charm.py
i-chvets: We would need to block charm if relation does not exist, right?

i-chvets:
> and another point, I do not see us monitoring the event if database endpoint changes
> 
> also, we need to put charm into Blocked state if DB relation is removed

When we put charm in Blocked state, is `update-status` still processed?

src/charm.py
misohu: yes we do it ... mlflow cant start without relational db

src/charm.py
```diff
@@ -149,25 +159,28 @@ def _get_interfaces(self):
...
-            db_unit = list(relational_db.units)[0]
-            relational_db = relational_db.data[db_unit]
-            relational_db["database"]
-            return relational_db
-        except IndexError:
-            raise ErrorWithStatus("Waiting for relational-db relation data", WaitingStatus)
-        except KeyError:
-            raise ErrorWithStatus("Missing data in relational-db relation", WaitingStatus)
+    def _get_relational_db_data(self) -> dict:
+        mysql_relation = self.model.get_relation("relational-db")
```
i-chvets, line 163: What can we check here except for `None` to ensure the relation is established?
The flow here is to check relation to make sure this charm is related to MySQL charm, but retrieve data through library.

src/charm.py
```diff
@@ -149,25 +159,28 @@ def _get_interfaces(self):
...
+                continue
+            self.logger.info("New mysql database endpoint is %s", val["endpoints"])
+            host, port = val["endpoints"].split(":")
+            db_data = {
+                "host": host,
+                "port": port,
+                "username": val["username"],
+                "password": val["password"],
+            }
+            return db_data
```
i-chvets, line 182: Can we also add `raise ErrorWithStatus` for the case where we do not find required attributes?

src/charm.py
misohu: get_relation only returns if relation is established ... othervise its none for the relation

src/charm.py
misohu: Added handling for `KeyError` exception + added unit test.

i-chvets:
LGTM.

beliaev-maksim (approved):

PR #128, opened 2023-04-19
KF-1003 Remove log message

misohu:
* Remove log message in relational db handler function.


beliaev-maksim (approved):
why not have it :)
but is OK

PR #132, opened 2023-04-20
KF-1005 fix relation error

misohu:
- Check for minio connection exception (add waiting status when occured)
- Add update_status event handler to recover while waiting for minio


.github/workflows/integrate.yaml
```diff
@@ -63,7 +63,7 @@ jobs:
         provider: microk8s
         channel: 1.24/stable
         # Pinned until this bug is resolved: https://bugs.launchpad.net/juju/+bug/1992833
-        bootstrap-options: "--agent-version=2.9.34"
+        bootstrap-options: "--agent-version=2.9.34 --model-default update-status-hook-interval='30s'"
```
i-chvets, line 66: I understand we decrease update interval to speed up test. My concern about not using standard, real-life timing of 5 minutes comes from MySQL integration work. In that case, workload that talks to the DB was started without DB config or with incorrect DB config and timed out at 2-3 minutes and update status at 5 mins interval properly caught this situation. In this case, can we confirm workload behaviour: timeout is not happening/crash is not happening beyond 30 seconds interval.

src/services/s3.py
```diff
@@ -41,6 +41,8 @@ def check_if_bucket_accessible(self, bucket_name):
             return True
         except botocore.exceptions.ClientError:
             return False
+        except botocore.exceptions.EndpointConnectionError as err:
+            raise err
```
ca-scribner, line 45: what does this addition do? iiuc, this catches and immediately re-raises the error, so why include it?

src/charm.py
```diff
@@ -56,6 +57,7 @@ def __init__(self, *args):
             self.framework.observe(self.on[rel].relation_changed, self._on_event)
         self._create_service()
 
+        self.framework.observe(self.on.update_status, self._on_event)
```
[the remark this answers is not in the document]
ca-scribner: We decided to do this for now so this is not buggy, but to talk about better ways in future.

ca-scribner (approved):

misohu:
Merging issue with update_status tracted in https://github.com/canonical/mlflow-operator/issues/133

PR #148, opened 2023-06-02
Added docs link

ColmBhandal:
Update metadata.yaml with docs link
Link is to new overview page in Discourse


misohu (approved):
Approving. We will add the links to tutorials later directly on discourse.

PR #151, opened 2023-06-06
Docs link bundle

ColmBhandal:
Updated bundle.yaml with link to new docs page

### DIFF
--- a/releases/latest/edge/mlflow/bundle.yaml
+++ b/releases/latest/edge/mlflow/bundle.yaml
@@ -1,5 +1,6 @@
 bundle: kubernetes
 name: mlflow
+docs: https://discourse.charmhub.io/t/mlflow-docs-index/10836
 applications:
   mlflow-minio:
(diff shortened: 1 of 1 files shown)


releases/latest/edge/mlflow/bundle.yaml
```diff
@@ -1,5 +1,6 @@
 bundle: kubernetes
 name: mlflow
+docs: https://discourse.charmhub.io/t/mlflow-docs-index/10836
```
kimwnasptd, line 3: The link looks OK, but let's update the structure after we merge this PR to align more with the initial gdoc we have for the Buldne docs

kimwnasptd (approved):

releases/latest/edge/mlflow/bundle.yaml
ColmBhandal: Good idea :+1:

PR #154, opened 2023-06-08
KF-2356 Use Rock

misohu:
Use rock instead of docker image

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -13,7 +13,7 @@ resources:
   oci-image:
     type: oci-image
     description: Backing OCI image
-    upstream-source: docker.io/ubuntu/mlflow:2.1.1_1.0-22.04
+    upstream-source: docker.io/charmedkubeflow/mlflow:v2.1.1_22.04_1
 requires:
   relational-db:
     interface: mysql_client


metadata.yaml
```diff
@@ -13,7 +13,7 @@ resources:
   oci-image:
     type: oci-image
     description: Backing OCI image
-    upstream-source: docker.io/ubuntu/mlflow:2.1.1_1.0-22.04
+    upstream-source: docker.io/charmedkubeflow/mlflow:v2.1.1_22.04_1
```
i-chvets, line 16: Was it manually published?

i-chvets:
Since we are integrating manually published ROCK, we need to provide at least some instructions on how it was published. This is until common CI for publishing gets approved (it's been in review for quite some time).

metadata.yaml
misohu: yes

misohu:
@i-chvets I have updated the spec with my example as we discussed https://docs.google.com/document/d/1wKd6lKGVnaPdwxSSHWqCZzjeCCB-p1UhSh_Vp6a6gng/edit#heading=h.w9czzrb1b4nw

DnPlas (approved):
LGTM, thanks @misohu

PR #155, opened 2023-06-08
Added Doc description

ColmBhandal:
Added proper doc description
Took couple of lines from Discourse

This is just a temporary measure
Excpect Discourse doc to override this when ready

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -2,7 +2,7 @@
 # See LICENSE file for licensing details.
 name: mlflow-server
 description: |
-  mlflow is a Kubernetes charm bundle for MLflow. This bundle automates the deployment and operation of MLflow on any Kubernetes cloud.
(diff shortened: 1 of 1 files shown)


metadata.yaml
```diff
@@ -2,7 +2,7 @@
 # See LICENSE file for licensing details.
 name: mlflow-server
 description: |
-  MLflow
+  mlflow is a Kubernetes charm bundle for MLflow. This bundle automates the deployment and operation of MLflow on any Kubernetes cloud.
```
beliaev-maksim, line 5: not sure about Cloud

K8s cluster maybe ?

misohu: Agree with @beliaev-maksim we can say **on any Kubernetes cluster backed by any public cloud**. Just wanted to say we have never tested it yet on **any public cloud**, we are just starting :D So far we are just testing on microk8s.

ColmBhandal: Makes sense. I kept it safe. Let me know if that works.

PR #162, opened 2023-06-21
KF-3620 custom Prometheus exporter 

misohu:
Integrate prometheus exporter with mlflow-server as second container.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -9,11 +9,17 @@ docs: https://discourse.charmhub.io/t/mlflow-server-docs-index/10816
 containers:
   mlflow-server:
     resource: oci-image
+  mlflow-prometheus-exporter:
+    resource: exporter-oci-image
 resources:
   oci-image:
     type: oci-image
     description: Backing OCI image
     upstream-source: docker.io/charmedkubeflow/mlflow:v2.1.1_22.04_1
+  exporter-oci-image:
+    type: oci-image
+    description: Image for Prometheus exporter
+    upstream-source: docker.io/charmedkubeflow/mlflow-prometheus-exporter:1.0-22.04
 requires:
   relational-db:
     interface: mysql_client
--- a/tests/integration/test_deploy_runners.py
+++ b/tests/integration/test_deploy_runners.py
@@ -17,7 +17,8 @@ class TestDeployRunners:
         """
         charm_under_test = await ops_test.build_charm(".")
         image_path = METADATA["resources"]["oci-image"]["upstream-source"]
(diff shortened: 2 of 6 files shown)


metadata.yaml
```diff
@@ -9,11 +9,17 @@ docs: https://discourse.charmhub.io/t/mlflow-server-docs-index/10816
...
   mlflow-server:
     resource: oci-image
+  mlflow-prometheus-exporter:
+    resource: exporter-oci-image
 resources:
   oci-image:
     type: oci-image
     description: Backing OCI image
     upstream-source: docker.io/charmedkubeflow/mlflow:v2.1.1_22.04_1
+  exporter-oci-image:
```
ca-scribner, line 19: The publish CI for this I think is failing because this is a new resource.  There are instructions [here](https://github.com/canonical/charming-actions/issues/58#issuecomment-1579282157)

tests/unit/test_operator.py
ca-scribner, line 1: Should we add a test here to show that, if the new container is not ready, the charm goes into Waiting?  Is there a test like that for the original container?

tests/integration/test_deploy_runners.py
ca-scribner, line 1: What can we add to the integration tests to prove the exporter is actually working?

ca-scribner (commented):
I think this looks good, but a few more tests would make that clearer

tests/unit/test_operator.py
```diff
@@ -451,12 +461,41 @@ def test_send_manifests(self, create_manifests: MagicMock, harness: Harness):
...
     @patch("charm.MlflowCharm._get_relational_db_data")
-    def test_on_event(
+    def test_on_event_wainting_for_exporter(
+        self,
+        get_relational_db_data: MagicMock,
+        get_object_storage_data: MagicMock,
+        harness: Harness,
+    ):
+        get_object_storage_data.return_value = OBJECT_STORAGE_DATA
+        get_relational_db_data.return_value = RELATIONAL_DB_DATA
```
beliaev-maksim, line 471: do not remember, can return value be set directly in the patch wrapper ?

tests/integration/test_deploy_runners.py
misohu: Added test for checking the EP for custom metrics

tests/unit/test_operator.py
misohu: You right. I refactored. Thanks :)

ca-scribner (approved):

PR #163, opened 2023-06-22
KF-3652 Set Renovate schedule

misohu:
* set schedule to Sunday only

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -46,6 +46,6 @@
         }
     ],
     "schedule": [
-        "on friday"
+        "on sunday"
     ]
 }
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -44,5 +44,8 @@
             ],
             "automerge": true
         }
+    ],
+    "schedule": [
+        "on friday"
```
beliaev-maksim, line 49: let's go for sunday, so, on Monday we get ready to go PR

misohu: Fixed

beliaev-maksim (approved):

PR #171, opened 2023-06-30
KF-3704 add bundle test in ci for eks aws

misohu:
* add ci for testing bundle on EKS (create eks cluster, deploy bundle, test mlflow connection, tear down cluster).
* using team's credential for tests 
* for now there is only manual trigger
* tested in my personal repo https://github.com/misohu/charmed-mlflow-eks-ci/actions/runs/5411831694/jobs/9835096296


.github/cluster.yaml
```diff
@@ -0,0 +1,41 @@
...
+      ebs: true
+  instanceType: t2.2xlarge
+  labels:
+    alpha.eksctl.io/cluster-name: scrumptious-wardrobe-1684842095
+    alpha.eksctl.io/nodegroup-name: ng-d06bd84e
+  maxSize: 2
+  minSize: 2
+  name: ng-d06bd84e
+  releaseVersion: ""
+  ssh:
```
orfeas-k, line 31: Couldn't we set `ssh.allow` to false? Do we have a `dektop-eu-central-1` key in the kubeflow-team AWS account?

.github/cluster.yaml
```diff
@@ -0,0 +1,41 @@
...
+    allow: true
+    publicKeyName: dektop-eu-central-1
+  tags:
+    alpha.eksctl.io/nodegroup-name: ng-d06bd84e
+    alpha.eksctl.io/nodegroup-type: managed
+  volumeSize: 100
+metadata:
+  name: kubeflow-test
+  region: eu-central-1
+  version: "1.24"
```
orfeas-k, line 41: Also, what do you think about using a newer version if MLFlow works with newer versions of Kubernetes?

orfeas-k: Although, if we have manually deployed MLFlow mainly on `1.24`, I sugged we should stick to it here as well.

.github/workflows/deploy-eks.yaml
```diff
@@ -0,0 +1,71 @@
...
+      
+      - name: Create cluster
+        run: |
+          eksctl create cluster -f .github/cluster.yaml
+          kubectl get nodes
+      
+      - name: Setup juju
+        run: |
+          juju add-k8s kubeflow --client
+          juju bootstrap --no-gui kubeflow kubeflow-controller
```
orfeas-k, line 60: Do we really need `--no-gui` option? I think not, but if you prefer to keep it, it's fine also.

.github/workflows/deploy-eks.yaml
```diff
@@ -0,0 +1,71 @@
...
+          juju add-model kubeflow
+      
+      - name: Test bundle deployment
+        run: |
+          tox -vve bundle-test -- --model kubeflow --keep-models -vv -s
+      
+      - name: Remove eks 
+        if: always()
+        run: |
+          eksctl delete cluster --name=kubeflow-test
```
orfeas-k, line 70: I 'll just note that we should definitely have a script for deleting volumes since `eksctl delete cluster` doesn't always take care of this. I 'm good with merging this as is in order to move forward, but I 'd prefer if we already had it. Also, we should be explicit (maybe add a TODO here or a task in Jira?) that we should not schedule this action without having automated the volumes deletion. This and mainly the how we will do it is up for discussion in the `CKF to EKS` spec so we can discuss further there.

tests/integration/test_bundle.py
```diff
@@ -0,0 +1,46 @@
...
+            status="active",
+            raise_on_blocked=False,
+            raise_on_error=False,
+            timeout=1500,
+        )
+
+    @retry(stop=stop_after_delay(5), wait=wait_fixed(1))
+    @pytest.mark.abort_on_fail
+    async def test_mlflow_connetion(self, forward_connections, ops_test: OpsTest):
+        mlflow_response = requests.get("http://localhost:5002")
```
orfeas-k, line 42: I 'm not sure I get why we have to `kubectl port-forward` here.

orfeas-k (commented):
I 'll let mainly @ca-scribner comment on the pytest part of the PR. The logic of the test though is pretty straightforward and looks good. Other than that, good job, this is a great contribution for deploying charms to EKS and testing them in general.

The only thing I would like to bring up here is the action scheduling. Right now, this is triggered manually. I 'm fine with merging it like this and testing it (since now we cannot see this running) but I think we should:

1. Add a Jira task for scheduling this action (in order to make sure we won't miss it)
2. Ensure volumes deletion is in place before scheduling this action, as stated in a review comment for `deploy-eks.yaml` file.

I 'll cc @kimwnasptd also just for him to be in sync with this since the three of us were discussing this live.

.github/workflows/deploy-eks.yaml
```diff
@@ -0,0 +1,71 @@
...
+          sudo apt-get update
+          sudo apt-get install -y unzip
+          curl --silent --location "https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_$(uname -s)_amd64.tar.gz" | tar xz -C /tmp
+          sudo mv /tmp/eksctl /usr/local/bin
+          eksctl version
+      
+      - name: Install juju
+        run: |
+          sudo snap install juju --classic
+          sudo snap install charmcraft --classic
```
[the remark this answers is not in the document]
beliaev-maksim: it should be used for charm packing in tests

.github/cluster.yaml
```diff
@@ -0,0 +1,41 @@
...
+  releaseVersion: ""
+  ssh:
+    allow: true
+    publicKeyName: dektop-eu-central-1
+  tags:
+    alpha.eksctl.io/nodegroup-name: ng-d06bd84e
+    alpha.eksctl.io/nodegroup-type: managed
+  volumeSize: 100
+metadata:
+  name: kubeflow-test
```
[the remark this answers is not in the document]
ca-scribner: I know nothing about anything here :)  If there's a potential name collision, what happens if I push two commits within a few minutes of each other?

.github/workflows/deploy-eks.yaml
```diff
@@ -0,0 +1,71 @@
...
+on:
+  workflow_dispatch: # This event allows manual triggering from the GitHub UI
+    secrets:
+      AWS_ACCESS_KEY_ID:
+        required: true
+      AWS_SECRET_ACCESS_KEY:
+        required: true
+jobs:
+  deploy-eks:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
ca-scribner: +1 to this suggestion.  We pin it (we used to have `ubuntu-latest`, but then they updated what latest meant last year and it broke our CI :) )

tests/integration/test_bundle.py
```diff
@@ -0,0 +1,46 @@
...
+    yield
+    mlflow_process.terminate()
+    exporter_process.terminate()
+
+
+class TestCharm:
+    @pytest.mark.abort_on_fail
+    async def test_deploy_bundle_works(self, ops_test: OpsTest):
+        subprocess.Popen(["juju", "deploy", f"{BUNDLE_PATH}", "--trust"])
+        ops_test.deploy_bundle(BUNDLE_PATH)
```
ca-scribner, line 30: should both of these be here?

tests/integration/test_bundle.py
orfeas-k: Yeah, this actually is the way we did it also while developing WAs if we needed a component and didn't want to build it locally. It looks like yesterday I was a bit tired to understand what was happening here :P . Anyway, the only thought I have about this is if this is a proper way to test the connection to MLFlow, since we wouldn't `port-forward` anything during a normal deployment. Wouldn't it make more sense to try and hit its public URL?

.github/cluster.yaml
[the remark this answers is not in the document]
orfeas-k: These are not triggered on push, but still, since they are long tests,IMHO there could be a name collision.

kimwnasptd:
Regarding deleting the EBS Volumes and next steps:
* Yes let's have a task for this. I'm OK if it's also a workflow we trigger manually (just deleting volumes) but let's still have something
* @orfeas-k let's have some of this logic on your spec as well

It could be something simple like delete all volumes that start with `kubeflow-test` and are `available` https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-deleting-volume.html

.github/workflows/deploy-eks.yaml
```diff
@@ -0,0 +1,71 @@
...
+          juju version
+      
+      - name: Create cluster
+        run: |
+          eksctl create cluster -f .github/cluster.yaml
+          kubectl get nodes
+      
+      - name: Setup juju
+        run: |
+          juju add-k8s kubeflow --client
```
orfeas-k, line 59: We could add `--eks` option to `juju add-k8s` command but no strong preference really since they both work fine.

.github/workflows/deploy-eks.yaml
```diff
@@ -0,0 +1,71 @@
...
+          juju bootstrap --no-gui kubeflow kubeflow-controller
+          juju add-model kubeflow
+      
+      - name: Test bundle deployment
+        run: |
+          tox -vve bundle-test -- --model kubeflow --keep-models -vv -s
+      
+      - name: Remove eks 
+        if: always()
```
orfeas-k, line 68: Also, I 'm not sure why we need the `if: always()` here.

.github/workflows/deploy-eks.yaml
orfeas-k: Let's drop this suggestion since `--eks` makes Juju ask for the region of the cluster while without it resolves this by itself.

.github/workflows/deploy-eks.yaml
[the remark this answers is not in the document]
orfeas-k: @beliaev-maksim Doesn't it pull all charms for this bundle test?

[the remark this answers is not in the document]
beliaev-maksim: @orfeas-k yes, I think that should be the way to test. Then we indeed do not need it

.github/cluster.yaml
misohu: I removed that section no need

.github/workflows/deploy-eks.yaml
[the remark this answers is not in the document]
misohu: We need it to run the pytest operator tests

.github/cluster.yaml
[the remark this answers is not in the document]
misohu: We will run only bi weekly. SO i dont think there will be a colision. When we decide to run more often then we must change it.

tests/integration/test_bundle.py
misohu: juju public url is juju think I wanted to test it more kubernetes like ... with port forward I can be 100% sure I can talk to it from cluster

tests/integration/test_bundle.py
misohu: SO I am doing retry for the later test If I connect them the retry will try do redeploy the bundle every time

.github/workflows/deploy-eks.yaml
misohu: As Kimonas mentioned bellow we will have a separate task for it. We can fix it there I will open issue in this repo

.github/workflows/deploy-eks.yaml
misohu: If you put always it will always run even if previous tasks fail.

.github/workflows/deploy-eks.yaml
misohu: I mean it is preferable ... why do I need gui if I dont access it :)

.github/workflows/deploy-eks.yaml
misohu: Here is issue https://github.com/canonical/mlflow-operator/issues/172

tests/integration/test_bundle.py
orfeas-k: Okk, commenting just to add the context I missed in order to understand the situation here. At the time of writing, when deploying MLFlow standalone (without Kubeflow), there is no Istio integration or LoadBalancer support to expose MLFlow.

orfeas-k:
So everything looks good to me (left one comment unresolved for @ca-scribner to take a look). Discussing it with @kimwnasptd , it would make sense to actually expose logs in case of failure. It could something like what we do [here](https://github.com/canonical/seldon-core-operator/blob/main/.github/workflows/integrate.yaml#L66-L80). WDYT @misohu ?

misohu:
@orfeas-k agree added to the CI

orfeas-k (approved):

ca-scribner (approved):

PR #179, opened 2023-07-18
Pass value of relation string directly

misohu:
Looks like we cant use environment variables in the command section of the layer.

* Pass value of mysql connection string directly to command
* Use ubuntu mlflow image

Please First merge this PR https://github.com/canonical/base-mlflow/pull/10

Comments to the Integration tests 
- I had to comment out the action test as the python function is not working after update of package juju==2.9.44 (manual test still works). I need that specific version for pyyayml problem with 6.0.0
- I had to rearrange order of tests. Looks like there have been some change in github runner disk space which cases to mysql charm to fail after some tests (all tests were kept just rearranged.)


.github/workflows/integrate.yaml
```diff
@@ -69,6 +69,11 @@ jobs:
       run: sg microk8s -c "tox -vve integration -- --model testing"
 
     # On failure, capture debugging resources
+    - name: Setup tmate session
+      uses: mxschmitt/action-tmate@v3
+      timeout-minutes: 60
+      if: failure()
+
```
orfeas-k, line 76:
```suggestion
```
Do we need this? Shouldn't we remove it?

tests/integration/test_charm.py
```diff
@@ -385,53 +420,23 @@ async def test_mlflow_alert_rules(self, ops_test: OpsTest):
...
-        config = await ops_test.model.applications[CHARM_NAME].get_config()
-        url = f"http://localhost:{config['mlflow_prometheus_exporter_nodeport']['value']}/metrics"
-        response = requests.get(url)
-        assert response.status_code == 200
-        metrics_text = response.text
-        assert 'mlflow_metric{metric_name="num_experiments"} 2.0' in metrics_text
-        assert 'mlflow_metric{metric_name="num_registered_models"} 0.0' in metrics_text
-        assert 'mlflow_metric{metric_name="num_runs"} 0' in metrics_text
+    #     assert access_key == OBJECT_STORAGE_CONFIG["access-key"]
+    #     assert secret_access_key == OBJECT_STORAGE_CONFIG["secret-key"]
```
orfeas-k, line 434: Do we need to keep this as a comment? Is there a chance we 're using it again in the future?

orfeas-k (commented):
LGTM, left minor comments, once they're OK, I 'll approve.

orfeas-k (dismissed):

orfeas-k (approved):

PR #189, opened 2023-08-16
Run EKS CI biweekly and add remove volumes job

misohu:
* Run the eks CI biweekly 
* Remove volume at the end of ekc ci run 
* Set correct env variables in repo 

Successful test run in private repo here https://github.com/misohu/charmed-mlflow-eks-ci/actions/runs/5876696690

### DIFF
--- a/.github/workflows/deploy-eks.yaml
+++ b/.github/workflows/deploy-eks.yaml
@@ -1,29 +1,30 @@
-name: Configure EKS and run MLflow bundle test
-
+name: Create EKS cluster, deploy CKF and run bundle test
 on:
-  workflow_dispatch: # This event allows manual triggering from the GitHub UI
+  workflow_dispatch: # This event allows manual triggering from the Github UI
     secrets:
-      AWS_ACCESS_KEY_ID:
+      BUNDLE_KUBEFLOW_EKS_AWS_ACCESS_KEY_ID:
         required: true
-      AWS_SECRET_ACCESS_KEY:
+      BUNDLE_KUBEFLOW_EKS_AWS_SECRET_ACCESS_KEY:
         required: true
+  schedule:
+    - cron: "23 0 * * 2"
 jobs:
-  deploy-eks:
+  deploy-ckf-to-eks:
     runs-on: ubuntu-22.04
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
-      
-      - name: Install tox and run tests
+
+      - name: Install tox
         run: |
-          sudo apt-get install -y python3-pip
-          sudo pip3 install tox
+          python -m pip install --upgrade pip
+          pip install tox
       
       - name: Configure AWS Credentials
         env:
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-eks.yaml
```diff
@@ -85,3 +86,11 @@ jobs:
       - name: Get operator logs
         run: kubectl logs --tail 100 -ntesting -loperator.juju.is/name=mlflow-server
         if: failure()
+
+  delete-unattached-volumes:
+    if: always()
+    uses: ./.github/workflows/delete-aws-volumes.yaml
+    secrets: inherit
+    with:
+      region: eu-central-1
```
kimwnasptd, line 95: Could we define this as an ENV var that gets shared across the whole YAML? I see that other steps are using this as well

.github/workflows/deploy-eks.yaml
```diff
@@ -1,37 +1,38 @@
-name: Configure EKS and run MLflow bundle test
-
+name: Create EKS cluster, deploy CKF and run bundle test
```
orfeas-k, line 1: deploy CKF or MLFlow?

.github/workflows/deploy-eks.yaml
```diff
@@ -1,37 +1,38 @@
...
+      BUNDLE_KUBEFLOW_EKS_AWS_ACCESS_KEY_ID:
         required: true
-      AWS_SECRET_ACCESS_KEY:
+      BUNDLE_KUBEFLOW_EKS_AWS_SECRET_ACCESS_KEY:
         required: true
+  schedule:
+    - cron: "23 0 * * 2"
 jobs:
-  deploy-eks:
+  deploy-ckf-to-eks:
```
orfeas-k, line 12: deploy ckf or mlflow?

.github/workflows/deploy-eks.yaml
```diff
@@ -85,3 +86,11 @@ jobs:
...
         run: kubectl logs --tail 100 -ntesting -loperator.juju.is/name=mlflow-server
         if: failure()
+
+  delete-unattached-volumes:
+    if: always()
+    uses: ./.github/workflows/delete-aws-volumes.yaml
+    secrets: inherit
+    with:
+      region: eu-central-1
+    needs: [deploy-ckf-to-eks]
```
orfeas-k, line 96: is `needs` necessary? won't this always run after the `deploy-ckf-to-eks` action?

orfeas-k (commented):
Thank you @misohu, left some minor comments.

.github/workflows/deploy-eks.yaml
misohu: I have nothing against the env, but would you mind if I dont do that  just to stay consistent with the kubeflow-bundle ci ?

kimwnasptd: @misohu that means that we keep reproducing code we know we don't want, just to stay consistent with code we don't want :)

So no, let's fix this

kimwnasptd:
The changes look good to me, so if @orfeas-k is also OK we should be good to merge this one

PR #190, opened 2023-08-16
fix: Pass `ENV_AUTH` value as string

phoevos:
According to the [Seldon docs](https://docs.seldon.io/projects/seldon-core/en/v1.14.0/servers/overview.html#example-minio-configuration), `RCLONE_CONFIG_S3_ENV_AUTH` receives a boolean value. However, passing `true` or `false` will cause the metacontroller to break when attempting to convert the Secret from JSON to a Go struct, since it's unable to unmarshall the boolean field. This in turn, will lead to the secret not being applied.

Consequently, we have to quote the boolean value to ensure it's interpreted as a string, e.g.
```yaml
RCLONE_CONFIG_S3_ENV_AUTH: "false"
```


i-chvets:
What a nasty bug.
Approved.

i-chvets (approved):
Approved.

PR #192, opened 2023-08-25
Fix integration tests to be executable on any Kubernetes cluster

misohu:
Fix tests to run on any Kubernetes. 

- Use port forward to access applications.
- Use hostname for loadbalancer if ip is not available 

Tested on EKS.


tests/integration/test_charm.py
```diff
@@ -204,95 +210,89 @@ async def test_add_relational_db_with_relation_expect_active(self, ops_test: Ops
     @pytest.mark.abort_on_fail
     async def test_can_connect_exporter_and_get_metrics(self, ops_test: OpsTest):
         config = await ops_test.model.applications[CHARM_NAME].get_config()
-        url = f"http://localhost:{config['mlflow_prometheus_exporter_nodeport']['value']}/metrics"
+        exporter_port = config["mlflow_prometheus_exporter_port"]["value"]
+        mlflow_subprocess = subprocess.Popen(
+            [
+                "kubectl",
+                "-n",
+                "kubeflow",
```
i-chvets, line 218: @misohu Looks like integration tests are failing. Is it because of the namespace we specify here?

tests/integration/test_charm.py
```diff
@@ -382,9 +382,19 @@ async def test_mlflow_alert_rules(self, ops_test: OpsTest):
...
         )
 
-        status = await ops_test.model.get_status()
-        prometheus_units = status["applications"]["prometheus-k8s"]["units"]
-        prometheus_url = prometheus_units["prometheus-k8s/0"]["address"]
+        prometheus_subprocess = subprocess.Popen(
+            [
+                "kubectl",
+                "-n",
+                "kubeflow",
```
i-chvets, line 389: Same comment as above.

tests/integration/test_charm.py
```diff
@@ -204,95 +210,89 @@ async def test_add_relational_db_with_relation_expect_active(self, ops_test: Ops
...
+                "port-forward",
+                f"svc/{CHARM_NAME}",
+                f"{exporter_port}:{exporter_port}",
+            ]
+        )
+        time.sleep(10)  # Must wait for port-forward
```
i-chvets, line 224: Is it safe to assume 10 seconds is enought for port forward to be established?

i-chvets:
Integration tests in GH runners are executed using model `testing`: `tox -vve integration -- --model testing`
In our test in this PR, when we try to setup port-forward, we assume model `kubeflow` (see in-line comments).
If we use `ops_test.model_name` instead it will make tests more portable and with no disturbance to workflows. Unless mlflow requires `kubeflow` model, the the workflow needs to change to use `kubeflow` model.

tests/integration/test_charm.py
i-chvets: Resolved in online sync up. 10 seconds should be enough.

i-chvets (approved):
Need to revisit local integration testing. Commands for locat tests need to be updated or reviewed. Local tests do not pass.
Separate [issue](https://github.com/canonical/mlflow-operator/issues/193) is filed to track this.
Approved.

PR #241, opened 2024-06-18
ci: bump juju 3.1 -> 3.5 in track/2.1

DnPlas:
Part of canonical/bundle-kubeflow#859

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -67,7 +67,7 @@ jobs:
     - name: Setup operator environment
       uses: charmed-kubernetes/actions-operator@1.1.0
       with:
-        juju-channel: "3.1/stable"
+        juju-channel: "3.5/stable"
         provider: microk8s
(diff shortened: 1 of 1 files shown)


DnPlas:
The problem with the CI is that mysql-k8s 8.0/edge requires juju <3.5.0 or ==3.5.2, which hasn't been released. This CI will be blocked until we do either of:

a) Use mysql-k8s 8.0/stable - this is problematic because they will soon promote from 8.0/edge, which will leave us with the same issue
b) Pin juju to a different version, it can be 3.4.3

orfeas-k:
I 'd go with a) since I 'd expect taking this requirement into account before promoting edge -> stable, since it's a breaking change in a way.

DnPlas:
https://github.com/canonical/mlflow-operator/pull/246 should help fix the CI. Lets merge that before this one.

DnPlas:
Closing this PR as it will be replaced by #241

Issue #266, opened 2024-09-09
Create a terraform module for mlflow-server charm


DnPlas:
### Context

According to the terraform modules organization spec, this is the first step to provide a terraform solution module. We need to create a terraform module for each charm.

### What needs to get done

Create a terraform module for this charm that complies with the [spec](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?usp=drivesdk).

### Definition of Done

Charm has a terraform module

PR #273, opened 2024-09-24
ci, docs: update to match standards for terraform

DnPlas:
This PR introduces the following changes based on a recent update to the team's standards for the terraform charm modules:

* docs: remove compatibility note

The latest standard for the terraform/ README.md is that the compatibility
    note can be obviated since the module at the branch is compatible with the
    charm in the same branch.

* ci: remove the channel from terraform checks

Since the default for deploying the charm is now set to latest/edge in
    the re-usable workflow, this is not required anymore.

Part of #266 (even though it was already closed)

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -55,7 +55,6 @@ jobs:
     uses: canonical/charmed-kubeflow-workflows/.github/workflows/terraform-checks.yaml@main
     with:
       charm-path: .
-      channel: latest/edge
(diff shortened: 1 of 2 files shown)


NohaIhab (commented):
changes lgtm, but the integration tests failed completely unrelated to the PR, I triggered a re-run to see if the issue persists. If it fails, let's file an issue and report it to DP team since it's caused by mysql operator.

NohaIhab (approved):
Tests passed with no issues, thx @DnPlas

Issue #287, opened 2024-10-16
Missing charm config to expose the --serve-artifacts option for mlflow


Barteus:
### Context

When accessing the mlflow from outside the cluster the S3 storage is not exposed, the default mlflow library in python requires flag `--serve-artifacts` to use tracking server to upload artefacts instead of directly uploading them to the object storage.

From mlflow documentation:

> If the MLflow server is not configured with the --serve-artifacts option, the client directly pushes artifacts to the artifact store. It does not proxy these through the tracking server by default.

For data security its better to not expose the S3 storage to the world if possible.

### What needs to get done

1. add config option
2. if server-artifacts == True, add flag `--serve-artifacts` to mlflow server execution

### Definition of Done

There is a way to enable the serve-artifacts flag for MLflow.

mathmarchand:
I can work on this, Can I be assigned to this issue?

PR #288, opened 2024-10-21
KF-6468 feat: Adding options to start server with artifacts options

mathmarchand:
Adding 2 options in config for:

- serve_artifacts: (boolean) add the switch to serve the artifacts of the mlflow server operations
- artifacts_destination: (string) location where the artifacts are to be stored.

Related to the issue #287 at the request of @Barteus

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -37,3 +37,13 @@ options:
     description: Deploy the NodePort service for MLFlow
     type: boolean
     default: true
+  serve_artifacts:
+    description: |
+      Enables serving of artifacts requests by routing these request to ./mlartifacts directory.
+    type: boolean
+    default: false
+  artifacts_destination:
+    description: |
+      Base location URI of resolve artifact requests.
+    type: string
+    default: './mlartifacts'
--- a/src/charm.py
+++ b/src/charm.py
@@ -50,6 +50,8 @@ class MlflowCharm(CharmBase):
 
         self.logger = logging.getLogger(__name__)
         self._port = self.model.config["mlflow_port"]
+        self._serve_artifacts = self.model.config.get("serve_artifacts", False)
+        self._artifacts_destination = self.model.config.get("artifacts_destination", "")
         self._exporter_port = self.model.config["mlflow_prometheus_exporter_port"]
         self._container_name = "mlflow-server"
         self._exporter_container_name = "mlflow-prometheus-exporter"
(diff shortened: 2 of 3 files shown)


misohu:
Thanks for working on adding artifact serving support! The `--serve-artifacts` feature is valuable for scenarios where clients can't directly access the object storage.

However, I've identified a critical issue with the current implementation:

## Problem

The PR introduces `artifacts_destination` with a default value of `'./mlartifacts'` (local ephemeral storage), but the existing charm already configures MLflow to use S3 for artifact storage via `default_artifact_root`.

Looking at [charm.py#L220](https://github.com/canonical/mlflow-operator/blob/main/src/charm.py#L220), the charm sets:
```bash
--default-artifact-root s3://{default_artifact_root}/
```

If we add `--artifacts-destination ./mlartifacts`, MLflow would:
1. Tell clients artifacts are at `s3://bucket/` (via `--default-artifact-root`)
2. Actually store proxied artifacts in `./mlartifacts` (local pod storage)
3. **Result: All artifacts would be lost on pod restart**

## Root Cause

The `--artifacts-destination` flag tells the MLflow server where to **actually store** artifacts when proxying, while `--default-artifact-root` tells clients where artifacts **should be located**. These must point to the same storage location.

## Suggested Solutions

**Option 1 (Recommended):** Make `artifacts_destination` automatically derive from `default_artifact_root`.

**Option 2:** Keep the separate `artifacts_destination` config but add validation to ensure it points to persistent storage and matches `default_artifact_root`.

---

The charm already handles S3 bucket creation and validation (see [`_create_default_s3_bucket`](https://github.com/canonical/mlflow-operator/blob/main/src/charm.py#L395-L407) and [`_validate_default_s3_bucket_name_and_access`](https://github.com/canonical/mlflow-operator/blob/main/src/charm.py#L410-L430)), so using the existing S3 configuration for artifact serving makes the most sense.

MattiaSarti:
@mathmarchand, I am working on more of a complete solution to resolve the same issue this pull request of yours is targeting, do you mind if I close it in favor of mine, once I have a working prototype?

mathmarchand:
> @mathmarchand, I am working on more of a complete solution to resolve the same issue this pull request of yours is targeting, do you mind if I close it in favor of mine, once I have a working prototype?

No problem at all!

MattiaSarti:
Closed in favor of https://github.com/canonical/mlflow-operator/pull/444

PR #306, opened 2025-03-17
chore(ci): bump GH runners to ubuntu-24.04

### LINKED ISSUE #305: Migrate CI from Ubuntu 20.04
### Context

Because of the [Ubuntu 20.04 deprecation](https://github.com/canonical/bundle-kubeflow/issues/1230) we'll need to move away from 20.04 in this and all CKF repos.

This will affect the CI, that is still relying on 20.04 and Python 3.8 for different actions (building, releasing, testing the charms).

Ideally we should migrate from 20.04 in all aspects of our charms (i.e. bases, python version etc) but this will be a significant undertaking. Thus for now we'll focus on:
1. Using the `actions/setup-python` to install Python 3.8 (affects global `pip` and `python3`). See example [here](https://github.com/canonical/microk8s/blob/0d619ff1c09dea97d9924a9604db263c639bcb50/.github/workflows/build-installer.yml#L20-L23)
2. Update the CI to use Ubuntu 24.04
3. Update the `publish.yaml` and/or `release.yaml` to install LXD and not use destructive mode, since we are now building on a different Ubuntu version than the host

### What needs to get done

1. Use the `actions/setup-python` to install Python 3.8, [example](https://github.com/canonical/microk8s/blob/0d619ff1c09dea97d9924a9604db263c639bcb50/.github/workflows/build-installer.yml#L20-L23)
2. Replace `runs-on: ubuntu-20.04` with `runs-on: ubuntu-24.04`, in the `.github/workflows`
3. Update the `publish.yaml` and/or `release.yaml` to install LXD and not use destructive mode, since we are now building on a different Ubuntu version than the host

### Definition of Done

1. The CI of the repo is not using Ubuntu 20.04 anywhere
2. The changes are merged to the `main` branch

orfeas-k:
Close #305

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   lib-check:
     name: Check libraries
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -24,28 +24,38 @@ jobs:
 
   lint:
     name: Lint Check
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     steps:
     - name: Check out code
       uses: actions/checkout@v2
 
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v5.3.0
+      with:
+        python-version: 3.8
+
     - name: Install dependencies
-      run: sudo apt-get install python3-pip tox
+      run: pip install tox
 
     - name: Lint code
       run: tox -e lint
 
   unit:
     name: Unit Test
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     steps:
     - name: Check out code
       uses: actions/checkout@v2
 
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v5.3.0
+      with:
+        python-version: 3.8
+
     - name: Install dependencies
-      run: sudo apt-get install python3-pip tox
+      run: pip install tox
 
     - name: Run unit tests
       run: tox -e unit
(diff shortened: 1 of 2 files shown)


mvlassis (commented):
Let's also update the version of the `checkout` action to  `v4` since we're updating the CI

.github/workflows/integrate.yaml
```diff
@@ -10,7 +10,7 @@ on:
 jobs:
   lib-check:
     name: Check libraries
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout
         uses: actions/checkout@v2
```
mvlassis, line 16:
```suggestion
        uses: actions/checkout@v4
```

.github/workflows/integrate.yaml
```diff
@@ -24,28 +24,38 @@ jobs:
 
   lint:
     name: Lint Check
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     steps:
     - name: Check out code
       uses: actions/checkout@v2
```
mvlassis, line 31:
```suggestion
      uses: actions/checkout@v4
```

.github/workflows/integrate.yaml
```diff
@@ -24,28 +24,38 @@ jobs:
...
       run: tox -e lint
 
   unit:
     name: Unit Test
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     steps:
     - name: Check out code
       uses: actions/checkout@v2
```
mvlassis, line 50:
```suggestion
      uses: actions/checkout@v4
```

mvlassis (commented):
I created suggestions as well

orfeas-k:
Updated

PR #321, opened 2025-05-09
refactor(tests): Move dependency charms to a separate file

orfeas-k:
Move dependency charms deployed during tests to a separate file called
charms_dependencies.py. This enables programmatic access to them and
thus updating them centrally, probably with the use of `kfcicli`. This
uses the `CharmSpec` dataclass from chisme.

Ref canonical/bundle-kubeflow/issues/1256


NohaIhab (approved):
thanks @orfeas-k

PR #361, opened 2025-07-31
[KF-7805] Disabling GH action for deploying to EKS

afgambin:
This PR addresses this Jira ticket: https://warthogs.atlassian.net/browse/KF-7805.

For review: I have renamed the files, instead of removing them, in case we want to keep them for future reference or re-enablement.

Also, do we want to remove related SECRETS to these two workflows from the repo?


afgambin:
Do we also want to remove the workflow for deploying to AWS, named `deploy-aws.yaml`?? Probably it makes sense in this case as well

afgambin:
Btw, I noticed the CI for this repo takes way longer than for others, such as bundle-kubeflow or charmed-kubeflow-solutions. Are we aware of this?

.github/workflows/deploy-eks.yaml.disabled
NohaIhab, line 1: > For review: I have renamed the files, instead of removing them, in case we want to keep them for future reference or re-enablement.

suggestion(non-blocking): since we are moving the workflows to the solutions repo I don't see a use for keeping this here. I suggest we just remove it.

NohaIhab (commented):
thanks @afgambin, replying to your comments:

> Do we also want to remove the workflow for deploying to AWS, named `deploy-aws.yaml`?? Probably it makes sense in this case as well

I see that workflow [has been failing](https://github.com/canonical/mlflow-operator/actions/workflows/deploy-aws.yaml)  for a while now, and we are not checking its results. I'm +1 to remove it but I'd do it in a separate PR, and we can first discuss with the team if sb thinks to fix it and keep it. 

> Btw, I noticed the CI for this repo takes way longer than for others, such as bundle-kubeflow or charmed-kubeflow-solutions. Are we aware of this?

yes, this is expected because the charm repos (like this one) run integration tests in the CI, while the other repos you touched before don't.

PR #385, opened 2025-11-27
feat: integrate with ambient mesh

### LINKED ISSUE #384: Integrate MLflow with ambient mesh
### Context

We'll also need to integrate MLflow with ambient mesh.

This component relies on Istio for now only for the Ingress relation, so that it gets exposed under the same ingress as CKF.

### What needs to get done

1. Add new relation logic for the new istio relation

### Definition of Done

1. Ensure MLflow can be deployed and function with Ambient Mesh charms and relatations

kimwnasptd:
Closes #384 

## Changes
1. Handle both sdi and mesh relations
2. Add unit tests for detecting if both relations exist

I'm putting this PR in draft, as we'll need to update it once we have the common code for testing the ingress redirection


MattiaSarti:
@NohaIhab, can we close this?

PR #428, opened 2026-06-19
fix: handle multiple `istio-ingress-route` relations 

NohaIhab:
Part of https://github.com/canonical/bundle-kubeflow/issues/1446

Use `model.relations[...]` instead of `model.get_relation(...)` when checking ingress relations, so the charm no longer raises `TooManyRelatedAppsError` when more than one `istio-ingress-route` relation is present.

Add unit tests covering multiple `istio-ingress-route` relations and verifying each relation receives a valid `HTTPRoute` config.

_Note: this PR was created with the help of AI_

### DIFF
--- a/tests/integration/test_charm_ambient.py
+++ b/tests/integration/test_charm_ambient.py
@@ -18,6 +18,8 @@ import requests
 import yaml
 from charmed_kubeflow_chisme.kubernetes import KubernetesResourceHandler
 from charmed_kubeflow_chisme.testing import (
+    ISTIO_INGRESS_K8S_APP,
+    ISTIO_INGRESS_ROUTE_ENDPOINT,
     assert_alert_rules,
     assert_grafana_dashboards,
     assert_logging,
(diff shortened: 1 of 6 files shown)


tests/integration/test_charm_ambient.py
MattiaSarti: nitpick: we could avoid repeating these same lines twice, as done in the other, analogous pull requests

tests/unit/test_operator.py
```diff
@@ -745,6 +749,84 @@ def test_istio_relations_conflict_detector(
...
+        add_relation(harness, relation_endpoint=RELATION_ENDPOINT_FOR_INGRESS_IN_AMBIENT_MODE)
+        second_app = f"app-for-{RELATION_ENDPOINT_FOR_INGRESS_IN_AMBIENT_MODE}-2"
+        second_relation_id = harness.add_relation(
+            RELATION_ENDPOINT_FOR_INGRESS_IN_AMBIENT_MODE, second_app
+        )
+        harness.add_relation_unit(second_relation_id, f"{second_app}/0")
+
+        # Inspecting the full list of relations per endpoint must not raise even
+        # though there is more than one relation on a single endpoint.
+        harness.charm._check_no_conflicting_ingress_relations()
```
MattiaSarti: suggestion (non-blocking): in addition to this, we could directly check the unit status, instead of using part of the charm (and hence of the tested logic) to perform assertions, which introduces bias

MattiaSarti (approved):

tests/unit/test_operator.py
dariofaccin: I think it's fine to check this instead of the unit status: the status might be `Blocked` (or anything) for other reasons, requiring other mocks and/or configurations.

dariofaccin (approved):

tests/unit/test_operator.py
MattiaSarti: You are right, @dariofaccin, but at the same time we are still using part of the tested logic to perform the test itself... yet I am fine with leaving this as is if you both are

dariofaccin: To my perspective this is a unit test of `_check_no_conflicting_ingress_relations()`. The issue is that we cannot test it without setting up the `harness`.

NohaIhab: I think Mattia has a good point on bias, it does take a lot of mocking but it's ok imo to make the test better

MattiaSarti (approved):

tests/unit/test_operator.py
MattiaSarti: Thanks, @NohaIhab!

PR #439, opened 2026-06-29
feat: add s3-credentials relation to relate with s3-integrator

### LINKED ISSUE #411: Add `s3` interface to relate with `s3-integrator`
### Context

.

### What needs to get done

.

### Definition of Done

.

dariofaccin:
## Overview

This PR adds support for the `s3` interface via a new `s3-credentials` relation,
allowing MLflow to use an external, generic S3-compatible object store (e.g.
AWS S3 through [`s3-integrator`](https://charmhub.io/s3-integrator)) as its
default artifact root — in addition to the existing `object-storage` (MinIO)
relation.

The charm now accepts **exactly one** of the `object-storage` or
`s3-credentials` relations and normalizes both into a common object-storage data
model used throughout the charm.

This PR closes #411.

### Charm (`src/charm.py`)
- Added the `S3Requirer` from the `data_platform_libs` object-storage library and
  observed `relation_changed` / `relation_broken` events for both
  `object-storage` and `s3-credentials`.
- Introduced `_get_object_storage_data()` which:
  - Errors (`Blocked`) if **both** relations are present.
  - Errors (`Blocked`) if **neither** relation is present.
  - Returns a normalized dict (`access-key`, `secret-key`, `host`, `port`,
    `secure`, `region`, `bucket`, `tls-ca-chain`, `is_s3`) for either relation.
- Added `_get_s3_data()` and `_parse_s3_endpoint()` to extract and validate the
  S3 connection info (host/port/TLS) from the `endpoint` provided by
  `s3-integrator`.
- Added a generic `_validate_sdi_interface()` helper for SDI relation validation
  with clearer `Blocked`/`Waiting` statuses.
- Added `_resolve_bucket_name()` which resolves the bucket from the relation
  data, falling back to the `default_artifact_root` config option.
- Replaced the old validate/create bucket flow with `_ensure_bucket_exists()`,
  which always creates the artifact-root bucket if missing and surfaces clearer
  statuses for TLS failures (`Blocked`) and connectivity issues (`Waiting`).
- Moved pebble layer/env generation to `_generate_environment` /
  `_mlflow_server_layer` / `_mlflow_exporter_layer` and switched to chisme's
  shared `update_layer` helper (removed the local `_update_layer`).
- Secrets/poddefaults contexts now derive the S3 endpoint scheme (`http`/`https`)
  from the relation's `secure` flag.

### S3 service (`src/services/s3.py`)
- Reworked `S3BucketWrapper` to support secure (HTTPS) endpoints, a configurable
  region (for `CreateBucketConfiguration` location constraint), and a custom
  TLS CA chain (written to a temporary PEM file with best-effort cleanup).
- Added connect/read timeouts via `botocore` `Config`.
- Replaced `check_if_bucket_accessible` / `create_bucket_if_missing` with
  `bucket_exists()`, `create_bucket()` (idempotent), and `delete_bucket()`.
- Removed the unused `validate_s3_bucket_name` helper.

### Metadata & config
- `metadata.yaml`: added the `s3-credentials` (`interface: s3`, `optional`,
  `limit: 1`) relation and marked `object-storage` as `optional` with `limit: 1`.
- `config.yaml`: deprecated `create_default_artifact_root_if_missing` — the charm
  now always creates the default artifact root bucket if it is missing.
- Removed the now-unused `src/secrets/mlflow-seldon-rclone-secret.j2` secret
  manifest.

### Tests & CI
- Renamed `tests/integration/test_charm.py` to
  `tests/integration/test_charm_object_storage.py` and added
  `tests/integration/test_charm_s3.py` covering the new `s3-credentials` relation.
- Added the `s3-integrator` (`2/edge`) charm dependency.
- Added an `integration-s3` tox environment and a corresponding
  `integration-s3` matrix entry to the CI workflow.
- Updated unit tests (`test_operator.py`, `test_s3.py`) for the new data model
  and `S3BucketWrapper` API.

## Testing

- `tox -e unit`
- `tox -e integration` (object-storage / MinIO)
- `tox -e integration-s3` (s3-credentials / s3-integrator)

## Backwards compatibility

- Existing deployments using the `object-storage` (MinIO) relation continue to
  work unchanged.
- The `create_default_artifact_root_if_missing` config option is now deprecated
  and has no effect (the bucket is always created if missing).


.github/workflows/ci.yaml
```diff
@@ -129,6 +129,7 @@ jobs:
           - mlflow-server
         test-type:
           - integration
```
mvlassis: nitpick: For consistency and clarity, let's rename this suit of tests to `integration-object-storage`. The understanding is that this will be deprecated and removed in the future. (Make sure to also change the `tox.ini`)

src/services/s3.py
mvlassis, line 1: note: This rewrite is based on https://github.com/canonical/kfp-operators/blob/0c3107e61f3e7758ff285a79a987a5bef6928891/charms/kfp-api/src/services/s3.py, which I agree with for consistency

tests/integration/test_charm_object_storage.py
```diff
@@ -436,3 +438,28 @@ async def test_new_user_namespace_has_manifests(
...
+        """Relating s3-integrator via s3-credentials should restore the charm to active."""
+        await deploy_and_assert_s3_integrator(
+            ops_test.model, s3_integrator=S3_INTEGRATOR, add_ca_chain=True
+        )
+        await ops_test.model.add_relation(
+            f"{CHARM_NAME}:s3-credentials", f"{S3_INTEGRATOR.charm}:s3-credentials"
+        )
+        await ops_test.model.wait_for_idle(
+            apps=[CHARM_NAME], status="active", raise_on_error=False, timeout=60 * 10
+        )
```
mvlassis, line 465: suggestion(non-blocking): The ideal thing to add here would be a test *after* the migration to test the workload. In the case of `mlflow-operator` that should probably be [test_new_user_namespace_has_manifests](https://github.com/canonical/mlflow-operator/blob/c4fa9cf4e8cd0ec307a55d26ce34341041b98a3f/tests/integration/test_charm.py#L421)

tests/integration/test_charm_s3.py
```diff
@@ -0,0 +1,387 @@
...
+        yaml_rendered["metadata"]["labels"][label] = "true"
+    obj = codecs.from_dict(yaml_rendered)
+    lightkube_client.apply(obj)
+
+    yield obj.metadata.name
+
+    delete_all_from_yaml(yaml_text, lightkube_client)
+
+
+async def setup_istio(ops_test: OpsTest, istio_gateway: CharmSpec, istio_pilot: CharmSpec):
```
mvlassis: suggestion(blocking): Since this feature is for any new deployments moving forward, we should rely on the new Ambient charms in the tests for s3. Let's use the same functions/fixtures from `test_charm_ambient.py`

tests/unit/test_s3.py
```diff
@@ -65,55 +56,94 @@ def s3_wrapper_empty():
...
-def test_check_if_bucket_accessible(
-    expected_returned, mocked_client, context_raised, s3_wrapper_empty
-):
+def test_bucket_exists(expected_returned, mocked_client, context_raised, s3_wrapper_empty):
     with context_raised:
         s3_wrapper_empty._client = mocked_client
 
-        bucket_name = "some_bucket"
-        returned = s3_wrapper_empty.check_if_bucket_accessible(bucket_name)
+        bucket_name = "some-bucket"
```
mvlassis, line 71: praise: Nice catch, the previous name is [not allowed in AWS](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html)

config.yaml
```diff
@@ -4,8 +4,8 @@
 options:
   create_default_artifact_root_if_missing:
     description: |
-      If True, charm will try to create the default_artifact_root bucket in S3 if it does not 
-      exist.  If False and the bucket does not exist, the charm enter Blocked status
+      DEPRECATED: this option is no longer used and has no effect. The charm now always
+      creates the default artifact root bucket in S3 if it does not exist.
```
mvlassis: note: We'll discuss with the team on how we should handle this

metadata.yaml
```diff
@@ -28,8 +28,14 @@ resources:
...
   relational-db:
     interface: mysql_client
+  s3-credentials:
+    interface: s3
+    optional: true
+    limit: 1
   object-storage:
     interface: object-storage
+    optional: true
+    limit: 1
```
mvlassis, line 38: note: This is backwards compatible. A CKF deployment with more than `object-storage` relations should *not* be allowed, and the expected behavior in `mlflow` is to [raise an error](https://github.com/canonical/mlflow-operator/blob/a9c1e6b43f7b3104efc3d002c6ba576b75b1264b/src/charm.py#L402-L409)

src/secrets/mlflow-seldon-rclone-secret.j2
mvlassis, line 1: question: What is this error? We should document why this is removed

metadata.yaml
MattiaSarti, line 38: note: similar considerations as in https://github.com/canonical/kserve-operators/pull/488#discussion_r3475570193 - all good to me

CONTRIBUTING.md
```diff
@@ -35,7 +35,8 @@ source .tox/unit/bin/activate
 ```shell
 tox -e lint          # code style
 tox -e unit          # unit tests
-tox -e integration   # integration tests
+tox -e integration-object-storage   # integration tests (object-storage relation)
+tox -e integration-s3               # integration tests (s3-credentials relation)
```
MattiaSarti: nitpick: as this point, we might as well:
```suggestion
tox -e integration-object-storage   # integration tests (sidecar mode & object-storage relation)
tox -e integration-ambient          # integration tests (ambient mode & object-storage relation)
tox -e integration-s3               # integration tests (ambient mode & s3-credentials relation)
```

src/services/s3.py
```diff
@@ -1,67 +1,110 @@
...
 
-        Will return False if we encounter a botocore.exceptions.ClientError, which could be
-        due to the bucket not existing, the client session not having permission to access the
-        bucket, or some other error with the client.
+    def create_bucket(self, bucket_name):
+        """Create a bucket."""
+        kwargs = {"Bucket": bucket_name}
+        # AWS S3 requires CreateBucketConfiguration for all regions except us-east-1, and
+        # rejects it for us-east-1. Use an empty region for non-AWS endpoints. See:
+        # https://docs.aws.amazon.com/boto3/latest/reference/services/s3/client/create_bucket.html#create-bucket
```
MattiaSarti, line 73: praise: helpful comment

src/charm.py
```diff
@@ -67,7 +76,6 @@
 ]
 SECRETS_FILES = [
     "src/secrets/mlflow-minio-artifact.j2",
-    "src/secrets/mlflow-seldon-rclone-secret.j2",
```
MattiaSarti, line 70: question: why is this not necessary anymore?

src/charm.py
MattiaSarti, line 126: suggestion (non-blocking): doesn't this already take care of lines 193 and 195? In fact, the object-storage relation had to be already observed before this pull request. I would avoid assigning event handlers twice with the new lines as it's error-prone, so maybe we can extend this old logic to `relation-broken` events, or change this old logic in favor of the new one

src/charm.py
```diff
@@ -352,6 +360,35 @@ def _mlflow_exporter_layer(self) -> Layer:
...
+        try:
+            interfaces = self._get_interfaces()
+            object_storage = self._get_object_storage_data(interfaces)
+        except ErrorWithStatus as error:
+            self.logger.error("Failed to generate container configuration.")
+            raise error
+        scheme = "https" if object_storage["secure"] else "http"
+        secrets_context = {
+            "app_name": self.app.name,
+            "s3_endpoint": f"{scheme}://{object_storage['host']}:{object_storage['port']}",
```
MattiaSarti: suggestion (non-blocking): without using `mlflow-seldon-rclone-secret.j2` anymore, can't we get rid of `s3_endpoint` as well? It's not used in `mlflow-minio-artifact.j2`, and we could code this directly in `poddefaults_context` instead of taking it from `secrets_context`

src/charm.py
```diff
@@ -389,32 +426,187 @@ def _get_relational_db_data(self) -> dict:
...
+            raise ErrorWithStatus("Waiting for s3-credentials relation data", WaitingStatus)
+        missing = [field for field in required_fields if not info.get(field)]
+        if missing:
+            raise ErrorWithStatus(
+                f"Waiting for s3-credentials relation data, missing fields: {', '.join(missing)}",
+                WaitingStatus,
+            )
+        return info
+
+    def _get_object_storage_data(self, interfaces=None) -> dict:
```
MattiaSarti: nitpick: method names are ambiguous, because `_get_object_storage_data` seems the counterpart of `_get_s3_data`, while its counterpart is actually `_get_object_storage`. What if, adopting MLflow's jargon, we refer to the "normalized", general one that encompassed both as "artifact store data", and keep "object storage data" as the counterpart for "S3 data"? We could always refer to the generalized storage data as the "artifact store data" throughout the source

src/charm.py
```diff
@@ -389,32 +426,187 @@ def _get_relational_db_data(self) -> dict:
...
 
-    def _on_get_minio_credentials(self, event):
+        if has_s3:
+            data = self._get_s3_data()
+            host, port, secure = self._parse_s3_endpoint(data["endpoint"])
+            if not host:
+                raise ErrorWithStatus(
+                    f"Invalid s3 endpoint: {data['endpoint']!r}",
+                    BlockedStatus,
+                )
```
MattiaSarti, line 577: nitpick: shouldn't this put the charm in waiting status, as the docstring says?

src/charm.py
```diff
@@ -389,32 +426,187 @@ def _get_relational_db_data(self) -> dict:
...
+            "tls-ca-chain": None,
+            "is_s3": False,
+        }
+
+    @staticmethod
+    def _parse_s3_endpoint(endpoint: str) -> tuple:
+        """Parse an s3 endpoint into a (host, port, secure) tuple.
+
+        The endpoint may be a full URL (e.g. "https://s3.example.com:443") or a bare
+        "host[:port]". kfp expects the host, port and TLS flag as separate values.
```
MattiaSarti: nitpick: no KFP here

src/charm.py
```diff
@@ -389,32 +426,187 @@ def _get_relational_db_data(self) -> dict:
...
+        return {
+            "access-key": obj["access-key"],
+            "secret-key": obj["secret-key"],
+            "host": f"{obj['service']}.{obj['namespace']}",
+            "port": obj["port"],
+            "secure": obj["secure"],
+            "region": "",
+            "bucket": "",
+            "tls-ca-chain": None,
+            "is_s3": False,
```
MattiaSarti: suggestion (non-blocking): how about a `TypedDict` not to hardcode keys twice?

src/charm.py
MattiaSarti, line 1: praise: thanks for tidying the source up along the way

tests/unit/test_s3.py
```diff
@@ -65,55 +56,94 @@ def s3_wrapper_empty():
...
-            None,
-            lf("client_accessible_emitting_unknown_exception"),
-            pytest.raises(Exception),
-        ),
+        # head_bucket succeeds -> the bucket exists
+        (True, lf("client_bucket_exists"), does_not_raise()),
+        # head_bucket returns a 404 -> the bucket does not exist (handled)
+        (False, lf("client_bucket_missing"), does_not_raise()),
+        # head_bucket returns any other error -> it propagates
+        (None, lf("client_bucket_forbidden"), pytest.raises(botocore.exceptions.ClientError)),
```
MattiaSarti, line 64: praise: well documented

tests/unit/test_operator.py
```diff
@@ -324,131 +340,141 @@ def test_get_relational_db_data_failure_waiting(self, harness: Harness):
...
+    def test_resolve_bucket_name_missing(self, harness: Harness):
+        """With no relation bucket and no config option, the charm blocks."""
+        harness.update_config({"default_artifact_root": ""})
         harness.begin()
-        value = harness.charm._validate_default_s3_bucket_name_and_access(BUCKET_NAME, s3_wrapper)
-        assert value
+        obj = {"bucket": "", "is_s3": False}
+        with pytest.raises(ErrorWithStatus) as exc_info:
+            harness.charm._resolve_bucket_name(obj)
+        assert exc_info.value.status_type(BlockedStatus)
```
MattiaSarti, line 372: suggestion (non-blocking): in place or in addition to this, can we also check the actual unit status? So that we also cover the propagation from raising `ErrorWithStatus` to actually setting the unit status

tests/unit/test_operator.py
```diff
@@ -324,131 +340,141 @@ def test_get_relational_db_data_failure_waiting(self, harness: Harness):
...
+            endpoint_url="http://minio.namespace:9000"
+        )
         harness.begin()
         with pytest.raises(ErrorWithStatus) as exc_info:
-            harness.charm._update_layer(container, harness.charm._container_name, MagicMock())
-
-        assert exc_info.value.status_type(BlockedStatus)
-        assert "Failed to replan with error: " in str(exc_info)
+            harness.charm._ensure_bucket_exists()
+        assert exc_info.value.status_type(WaitingStatus)
```
MattiaSarti, line 434: suggestion (non-blocking): same comment as above

src/services/s3.py
MattiaSarti, line 1: suggestion (blocking): I would mirror the original ambient-mode integration tests altogether, including the final `test_deploy_and_relate_second_ingress`, `test_httproute_attached_to_second_gateway` and `test_ui_is_accessible_after_second_ingress` too, in case in the future we had to remove MinIO-based object storage and wrongly assume that we could afford to delete the `test_charm_ambient.py` file because this new one has the same contents but migrated from MinIO to S3

MattiaSarti: (I commented on the wrong file by mistake, it was obviously for `tests/integration/test_charm_s3.py`)

src/charm.py
dariofaccin: I spoke with @misohu and we are not using Seldon anymore.

config.yaml
mvlassis: After discussion, we have decided to remove this config option. The changes here will be part of the new `3.x` track, so breaking backwards compatibility is acceptable.

MattiaSarti (approved):
Perfect!

note: before approving, I've removed the backport label as we agreed to include this in MLflow 3 only

note: while this is fine to me, I have a note about [the final commit that removes the deprecated config option](https://github.com/canonical/mlflow-operator/pull/439/commits/b42cd04ff6c457c72fc814c1996ad98503048b6a). Even if this will be included in MLflow 3 only, I believe we may want to keep the removed config option as deprecated during the next major release before removing it altogether, even if it has no effect on the charm logic anymore. @mvlassis, hadn't we discussed this during some past daily? Or am I hallucinating? Because now I see [this comment of yours](https://github.com/canonical/mlflow-operator/pull/439#discussion_r3505812192), but I remember this decision of keeping deprecated configs for a release.

src/charm.py
```diff
@@ -504,6 +738,28 @@ def _check_no_conflicting_ingress_relations(self) -> None:
...
+        See here how such environment variables provide defaults for `mlflow server` CLI options:
+        https://mlflow.org/docs/2.22.1/api_reference/cli.html#mlflow-server
+        """
+        try:
+            interfaces = self._get_interfaces()
+            object_storage = self._get_artifact_store_data(interfaces)
+            relational_db_data = self._get_relational_db_data()
+        except ErrorWithStatus as error:
+            self.logger.error("Failed to generate container configuration.")
+            raise error
```
mvlassis, line 753: question: I have a feeling we don't actually catch this error, don't we want to do so and emit a `BlockedStaus`?

src/charm.py
```diff
@@ -389,91 +447,267 @@ def _get_relational_db_data(self) -> dict:
...
+        Supports both the `object-storage` and `s3` interfaces, returning a common dict with
+        keys: access_key, secret_key, host, port, secure, region, bucket, tls_ca_chain, is_s3.
+
+        Exactly one of the `object-storage` or `s3-credentials` relations is expected.
+
+        Raises:
+            ErrorWithStatus(..., Blocked) if both relations are established at once.
+            ErrorWithStatus(..., Blocked) if neither relation is established.
+            ErrorWithStatus(..., Waiting) if the active relation has no data yet.
+        """
```
mvlassis, line 552: question: I like the raises here, but I think they are not caught either? We probably want to set the charm status too. In general I would safer if we actually unit test *all* or most of these errorwithstatus, to ensure they are actually set as the charm status

mvlassis (commented):
Just make sure we properly  handle the `ErrorWithStatus`

src/charm.py
MattiaSarti: Isn't is caught [in the final `_on_event` handler](https://github.com/canonical/mlflow-operator/blob/KF-8702-add-s3-interface/src/charm.py#L860)? Propagating like this: `_generate_environment` -> `service_environment` -> `_mlflow_server_layer` -> `update_layer` -> `_on_event`

mvlassis (approved):
Approving, the emitted `ErrorWithStatus` should be caught [here](https://github.com/canonical/mlflow-operator/blob/b42cd04ff6c457c72fc814c1996ad98503048b6a/src/charm.py#L860-L863)

PR #443, opened 2026-07-07
fix: explicitly refuse proxy mode

### LINKED ISSUE #442: Explicitly Refuse Proxy Mode
### Bug Description

The source versions of our charm does not employ the officially documented way to disable proxying artifacts through the tracking server, as it does not run the workload by setting either `--no-serve-artifacts` or `MLFLOW_SERVE_ARTIFACTS` - a behavior that is analogous both to the current version 2 and to version 3 - see:
- Page https://mlflow.org/docs/3.14.0/self-hosting/architecture/tracking-server/#tracking-server-no-proxy
- Subpage **_Other Configuration with MLflow Tracking Server_** -> **_Direct Access to Artifacts_**:
  - Target version docs: https://mlflow.org/docs/3.14.0/tracking/#other-tracking-setup
  - Source version docs: https://mlflow.org/docs/2.22.0/tracking/#other-tracking-setup

As a consequence, the tracking server does not refuse to act as a proxy for artifact upload, if explicitly asked to by the client. But by default artifacts are uploaded directly from the client side to the artifact store regardless, as long as the client does not override the default artifact root that is set (by our charm) on the server side, which is intentionally prefixed with `s3://` so that experiments have that in their default artifact URI and are therefore uploaded to the artifact store directly from the client by default, and not through the proxy (i.e., the tracking server). At the same time, though, this implies that clients should not modify this default behavior, as the tracking server is not configured to refuse to execute such requests, missing `--no-serve-artifacts` or `MLFLOW_SERVE_ARTIFACTS`, but it is not configured to connect to S3 either, so it would use its local artifact store on its filesystem in such a case, as neither `--artifacts-destination` nor `MLFLOW_ARTIFACTS_DESTINATION` are set:
- https://mlflow.org/docs/2.22.1/api_reference/cli.html#cmdoption-mlflow-server-artifacts-destination

Setting either `--no-serve-artifacts` or `MLFLOW_SERVE_ARTIFACTS` would instead ensure that, in this edge case, clients see their request fast-fail with the proper message returned by the server, with such a message being that the server is not enabled to proxy artifacts, instead of (silently) writing to the local server’s filesystem, which is not suitable for production (because of the replicated, ephemeral pod setup of the tracking server that does not make its filesystem a suitable persistence layer).

Therefore, `MLFLOW_SERVE_ARTIFACTS` can be added to the environment of the server, regardless of its backward-incompatible nature, given that it represents an undesired edge case where the charm is not used as intended if clients decide to override the default settings and let the tracking server proxy artifacts to then see them (erroneously) uploaded on its local filesystem.

### To Reproduce

N/A

### Environment

N/A

### Relevant Log Output

```shell
N/A
```

### Additional Context

_No response_

MattiaSarti:
resolves https://github.com/canonical/mlflow-operator/issues/442

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -758,6 +758,7 @@ class MlflowCharm(CharmBase):
             "MLFLOW_EXPOSE_PROMETHEUS": METRICS_PATH,
             "MLFLOW_HOST": "0.0.0.0",
             "MLFLOW_PORT": self._mlflow_port,
+            "MLFLOW_SERVE_ARTIFACTS": "False",
         }
 
     def _reconcile_policy_resource_manager(self):
--- a/tests/unit/test_operator.py
+++ b/tests/unit/test_operator.py
@@ -67,6 +67,7 @@ EXPECTED_ENVIRONMENT = {
     "MLFLOW_EXPOSE_PROMETHEUS": "/metrics",
     "MLFLOW_HOST": "0.0.0.0",
     "MLFLOW_PORT": 5000,
+    "MLFLOW_SERVE_ARTIFACTS": "False",
 }
 EXPECTED_SERVICE = {
     "mlflow-server": Service(


deusebio (approved):
Simple, yet effective! Thanks!

Backporting to 2.22 is very sensible to prevent people from using local storage to store models and risking to lose data!

misohu (approved):

PR #444, opened 2026-07-07
feat: parametrize proxy mode

### LINKED ISSUE #287: Missing charm config to expose the --serve-artifacts option for mlflow

MattiaSarti:
resolves https://github.com/canonical/mlflow-operator/issues/287

Also, in addition to parametrizing proxy mode, this pull request also addresses a lack of the previous integration with S3 storage where some necessary environment variables for the region and for the CA chain of S3 storage were not passed to clients (when in non-proxy mode) when those settings were passed as non-empty data by the S3 integration, in order not to see client requests fail in such cases.

To recap:
- when in proxy mode (config option `serve_artifcts` = `true`):
  - tracking server (re)run for direct S3 access
  - Secrets and PodDefaults updated to transfer S3 access to clients
- when not (default, as it always was in the past, config option `serve_artifcts` = `false`):
  - tracking server (re)run without S3 access
  - Secrets and PodDefaults updated to revoke S3 access from clients

note: **NOT** to be backported to MLflow 2, only available in MLflow 3

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -31,3 +31,11 @@ options:
     description: Deploy the NodePort service for MLFlow
     type: boolean
     default: true
+  serve_artifacts:
+    description: |
+      Whether to let the tracking server act as a proxy to interact with the artifact store on behalf of the client.
+
+      Analogous to:
+      https://mlflow.org/docs/latest/api_reference/cli.html#cmdoption-mlflow-server-serve-artifacts
+    type: boolean
+    default: false
--- a/src/charm.py
+++ b/src/charm.py
@@ -396,9 +396,7 @@ class MlflowCharm(CharmBase):
             # Base64-encoded S3 CA bundle so clients doing direct (non-proxy) object storage
             # I/O can mount it and trust the store's TLS certificate. Empty when there is none.
             "s3_ca_bundle_b64": (
-                base64.b64encode("\n".join(tls_ca_chain).encode()).decode()
-                if tls_ca_chain
-                else ""
+                base64.b64encode("\n".join(tls_ca_chain).encode()).decode() if tls_ca_chain else ""
             ),
         }
         return secrets_context
--- a/src/charm.py
+++ b/src/charm.py
@@ -763,23 +763,30 @@ class MlflowCharm(CharmBase):
             self.logger.error("Failed to generate container configuration.")
             raise error
 
+        s3_bucket_uri = f"s3://{self._resolve_bucket_name(artifact_store_data)}"
+
         environment_variables = {
             "MLFLOW_BACKEND_STORE_URI": f"mysql+pymysql://{relational_db_data['username']}:{relational_db_data['password']}@{relational_db_data['host']}:{relational_db_data['port']}/{self._database_name}",  # noqa: E501
-            "MLFLOW_DEFAULT_ARTIFACT_ROOT": f"s3://{self._resolve_bucket_name(artifact_store_data)}",
             "MLFLOW_EXPOSE_PROMETHEUS": METRICS_PATH,
             "MLFLOW_HOST": "0.0.0.0",
             "MLFLOW_PORT": self._mlflow_port,
-            "MLFLOW_SERVE_ARTIFACTS": "False",
         }
-
         if self.proxy_mode:
             environment_variables.update(
                 {
                     "AWS_ACCESS_KEY_ID": artifact_store_data["access_key"],
                     "AWS_SECRET_ACCESS_KEY": artifact_store_data["secret_key"],
+                    "MLFLOW_ARTIFACTS_DESTINATION": s3_bucket_uri,
+                    "MLFLOW_DEFAULT_ARTIFACT_ROOT": "mlflow-artifacts:/",
                     "MLFLOW_S3_ENDPOINT_URL": self._extract_s3_endpoint(artifact_store_data),
-                    # TODO: how about artifact destination?
                     "MLFLOW_SERVE_ARTIFACTS": "True",
+                }
+            )
+        else:
+            environment_variables.update(
+                {
+                    "MLFLOW_DEFAULT_ARTIFACT_ROOT": s3_bucket_uri,
+                    "MLFLOW_SERVE_ARTIFACTS": "False",
                 }
             )
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -80,8 +80,9 @@ SECRETS_FILES = [
     "src/secrets/mlflow-minio-artifact.j2",
 ]
 SERVICE_MESH_RELATION_NAME = "service-mesh"
-# Path inside the workload container where the artifact store's TLS CA bundle is written, so the
-# NOTE: under the Pebble user's home directory, writable by the non-root user:
+# path inside the workload container where the artifact store's TLS CA bundle is written to be then
+# referenced by the AWS_CA_BUNDLE environment variable, so that the tracking server can trust the
+# store's TLS certificate - NOTE: under Pebble's home directory, writable by the non-root user:
 S3_CA_BUNDLE_CONTAINER_PATH = "/var/lib/pebble/default/s3-ca-bundle.pem"
 
 
@@ -393,8 +394,8 @@ class MlflowCharm(CharmBase):
             "access_key": artifact_store_data["access_key"],
             "secret_access_key": artifact_store_data["secret_key"],
             "is_proxy_mode_enabled": self.proxy_mode,
-            # Base64-encoded S3 CA bundle so clients doing direct (non-proxy) object storage
-            # I/O can mount it and trust the store's TLS certificate. Empty when there is none.
+            # base64-encoded S3 CA bundle for clients to directly trust the artifact store when in
+            # no proxy mode and when private TLS certificates are used:
             "s3_ca_bundle_b64": (
                 base64.b64encode("\n".join(tls_ca_chain).encode()).decode() if tls_ca_chain else ""
             ),
@@ -417,8 +418,8 @@ class MlflowCharm(CharmBase):
                 f"{self._mlflow_port}"
             ),
             "is_proxy_mode_enabled": self.proxy_mode,
-            # Whether to mount the S3 CA bundle into client pods and point AWS_CA_BUNDLE at it,
-            # so their direct (non-proxy) boto3 connections trust the store's TLS certificate.
+            # whether to mount the S3 CA bundle into client pods and point AWS_CA_BUNDLE at it, so
+            # their direct (no-proxy mode) connections trust the artifact store's TLS certificate:
             "s3_ca_bundle_present": bool(artifact_store_data["tls_ca_chain"]),
         }
         return poddefaults_context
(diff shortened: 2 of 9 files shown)


MattiaSarti:
note: CI red due to flakiness, all tests passed before the last commit, which just updated code comments

config.yaml
```diff
@@ -31,3 +31,11 @@ options:
...
     type: boolean
     default: true
+  serve_artifacts:
+    description: |
+      Whether to let the tracking server act as a proxy to interact with the artifact store on behalf of the client.
+
+      Analogous to:
+      https://mlflow.org/docs/latest/api_reference/cli.html#cmdoption-mlflow-server-serve-artifacts
+    type: boolean
+    default: false
```
mvlassis, line 41: question(blocking): I can see that the default value for this, specified in the [documentation](https://mlflow.org/docs/latest/api_reference/cli.html#cmdoption-mlflow-server-serve-artifacts), is `true`. Perhaps we want the same for the operator as well?

MattiaSarti: I would advise against that because in our use case the charm has always been available in non-proxy mode, the equivalent of that to `False`, and that's why I kept it as the default values: so that upgrades from previous states with no config specified do not disrupt clients

src/charm.py
```diff
@@ -79,6 +80,10 @@
     "src/secrets/mlflow-minio-artifact.j2",
 ]
 SERVICE_MESH_RELATION_NAME = "service-mesh"
+# path inside the workload container where the artifact store's TLS CA bundle is written to be then
+# referenced by the AWS_CA_BUNDLE environment variable, so that the tracking server can trust the
+# store's TLS certificate - NOTE: under Pebble's home directory, writable by the non-root user:
+S3_CA_BUNDLE_CONTAINER_PATH = "/var/lib/pebble/default/s3-ca-bundle.pem"
```
deusebio, line 86: **question** I was also checking the image, but I didn't find permission to be able to write at this location. I'm not sure how we can. I read that this is taken as the Pebble home's directory, but I would have thought that we had to set it here: https://github.com/canonical/base-mlflow/blob/main/mlflow/rockcraft.yaml#L49

Just for the sake of my understanding, could you explain how this would work? Thanks!

config.yaml
deusebio: Yes, I believe this is fine and keep this to False to be consistent with previous deployments. We could change the default given that it is a major bump, but we can consider this later if we REALLY want to provide a stronger nudge towards this. For now, I believe it is actually simpler to keep it consistent with previous versions

src/charm.py
```diff
@@ -671,19 +696,19 @@ def _resolve_bucket_name(self, obj: dict) -> str:
 
     def _ensure_bucket_exists(self) -> None:
         """Ensure bucket on object storage exists by using a boto3 client."""
-        obj = self._get_artifact_store_data()
+        artifact_store_data = self._get_artifact_store_data()
```
deusebio, line 699: **issue,question** in this code base, although we use s3-integrator interface, it seems that we are not really using the object-storage library. Is there any reason not to?

src/charm.py
```diff
@@ -379,37 +384,51 @@ def _mlflow_exporter_layer(self) -> Layer:
     def secrets_context(self) -> dict:
         try:
             interfaces = self._get_interfaces()
-            object_storage = self._get_artifact_store_data(interfaces)
+            artifact_store_data = self._get_artifact_store_data(interfaces)
         except ErrorWithStatus as error:
             self.logger.error("Failed to generate container configuration.")
             raise error
+        tls_ca_chain = artifact_store_data["tls_ca_chain"]
```
deusebio, line 391: **praise** Adding support for TLS CA chain is indeed a good use-case we were possibly a bit short before, since if I understand the changes, this bits were not actually passed or accounted for. Nice!

src/poddefaults/poddefault-minio.yaml.j2
```diff
@@ -1,3 +1,4 @@
+{%- if not is_proxy_mode_enabled %}
```
deusebio, line 1: **thought** eventually, we will probably deprecate these things, since the MLFlow on 3/ will expose an client relation, and these resources will be created via kubeflow-integrator relations. But that's fine to keep this one as is and supporting this. I'm just saying that this will probably be removed here and be placed in data-kubeflow integrator. Just bear this in mind.

deusebio (commented):
Good job @MattiaSarti ! The code changes are quite sensible here. I have had to review a bit more closely the code to familiarize with it, but everything seems to be sensible and align with the feature request. The tests helped substantially to make sure that we were covering different cases. 

I have a few non-blocking comments, or comments that are out of scope. In particular, in the latter, I feel that we would probably be using the object-storage library to encode serialization/deserialization from the s3-credentials endpoint. But again, this is beyond the current task, and it can be approached as separate item (tech debt)

src/charm.py
MattiaSarti: Sure!

It's a suggestion we received from some external team member when reaching out to solve those mount problems, and we already adopted it in other repositories such as resource-dispatcher, where as you can see, [the charm](https://github.com/canonical/resource-dispatcher/blob/8b655320cd18bfef63aa754fd9905a2c1ed91268/src/charm.py#L31) relies on that [without the rock](https://github.com/canonical/resource-dispatcher-rock/blob/main/resource-dispatcher/rockcraft.yaml) being explicitly configured for permissions there.

From what I recall, it's [Pebble's home directory](https://ubuntu.com/docs/pebble/reference/environment-variables/#pebble) and by definition it is writable to the user the Pebble workload is run with, so that we don't have to modify the rock every time we touch the user ID in `metadata.yaml`.

I believe we could adopt this more and more in the future, but the suggestion came kind of late and most of our rocks were already adapted to add specific user permissions to specific locations - or had constrains on locations we had to write to that we could not change, but for use cases where we can write at any location, like in here, I would suggest always relying on this without changing rocks.

src/charm.py
MattiaSarti: Thanks!

Exactly, it used to fail when we had custom CA chains because they were not propagated (a use case we did not test in the CI).

src/charm.py
MattiaSarti: I'm not sure, let me tag @dariofaccin and @mvlassis, who worked on that right before this pull request and are probably more familiar with the object-storage library than I am (I wasn't aware of any such library).

src/charm.py
deusebio: Oh I see. Good to know! Thanks!

deusebio (approved):
Thanks for the feedback and responses! I'm fine with the changes here. Just sync with @theoctober19th and @dariofaccin around using the library for managing relation data and hooks in a standard manner, but this can be addressed separately from this PR

I don't need to re-review, unless there are substantial changes.

src/charm.py
mvlassis: We do use the object storage library, through the `S3Requirer` object, which uses the interface. See the [chisme module](https://github.com/canonical/charmed-kubeflow-chisme/blob/db3951093c0ab80719024debca0696577c550f4e/src/charmed_kubeflow_chisme/components/s3_component.py)

tests/integration/test_charm_ambient.py
```diff
@@ -161,6 +164,24 @@ async def profile_namespace(ops_test: OpsTest, lightkube_client: lightkube.Clien
         pass
 
 
+def _assert_resource_pruned(lightkube_client, resource, name: str, namespace: str):
```
mvlassis: thought: What do we mean by "pruning" here?. From what I understand, in the Kubernetes context pruning refers to a very specific `--prune` flag introduced for [kubectl](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#alternative-kubectl-apply-f-directory-prune).  

From what I see, this concept is not present either in [resource-dispatcher](https://github.com/search?q=repo%3Acanonical%2Fresource-dispatcher%20prune&type=code) or the [upstream metacontroller project](https://metacontroller.github.io/metacontroller/intro.html)

tests/integration/test_charm_ambient.py
```diff
@@ -654,3 +675,107 @@ async def test_ui_is_accessible_after_second_ingress(
...
+            [
+                "kubectl",
+                "-n",
+                f"{ops_test.model_name}",
+                "port-forward",
+                f"svc/{CHARM_NAME}",
+                f"{mlflow_port}:{mlflow_port}",
+            ]
+        )
+        time.sleep(10)  # Must wait for port-forward
```
mvlassis, line 745: suggestion(non-blocking): Can we use lightkube for this? (and thus avoid a `subprocess` call as well as an arbitrary `time.sleep` period

tests/integration/test_charm_object_storage.py
```diff
@@ -476,10 +510,142 @@ async def test_new_user_namespace_has_manifests_after_migration(
...
+    async def test_client_logs_and_fetches_artifact_via_tracking_server(self, ops_test: OpsTest):
+        """A client without object-storage access must round-trip artifacts in proxy mode.
+
+        With serve_artifacts enabled, artifacts are proxied through the tracking server, so a
+        client that only knows the tracking URI (and has no S3 credentials) must be able to
+        complete a full artifact round-trip.
+        """
+        config = await ops_test.model.applications[CHARM_NAME].get_config()
+        mlflow_port = config["mlflow_port"]["value"]
+        mlflow_subprocess = subprocess.Popen(
```
mvlassis, line 605: same as above regarding usage of `lightkube`

src/charm.py
mvlassis, line 86: nitpick: I would maybe add as a note that this makes the charm non-usable for a standard Docker image with a non-root user (due to permission issues)

mvlassis (approved):
Looks good! Left a few comments

tests/integration/test_charm_ambient.py
MattiaSarti: Good point, I simply mean that a previously resource is correctly cleared by resource dispatcher, and I've updated function names and docstrings accordingly

tests/integration/test_charm_ambient.py
MattiaSarti: I am afraid `lightkube` does not support port-forwarding, being just a declarative Kubernetes API client (get/list/create/apply/delete/watch on resources) and not implementing the SPDY/WebSocket streaming protocol that port-forwarding relies on. Indeed, we already used a similar piece of code on older parts of the test source, probably because an alternative was never available in the first place

tests/integration/test_charm_object_storage.py
MattiaSarti: (same reply as above)

src/charm.py
MattiaSarti: I don't think that's accurate, because (as far as I know, at least) Pebble creates and owns that directory regardless on the OCI image being run - it's not specific to rocks, it's something Pebble itself sets up for its own purposes in the workload when run as the entrypoint

Issue #432, opened 2026-06-22
Mlflow 3 Upgrade Documentation


MattiaSarti:
### Context

The charm is to be upgraded to MLflow 3.

### What needs to get done

Document the upgrade procedure for charm users.

### Definition of Done

The upgrade procedure is documented for charm users.

Issue #431, opened 2026-06-22
MLflow 3 Documentation Update


MattiaSarti:
### Context

The charm is to be upgraded to MLflow 3.

### What needs to get done

Update the documentation for MLflow 3, in case any changes are necessary from the user's perspective.

### Definition of Done

The documentation is updated for MLflow 3, in case any changes are necessary from the user's perspective.

PR #448, opened 2026-07-13
feat: upgrade to 3.14

### LINKED ISSUE #425: Upgrade Charm to MLflow 3
### Context

The charm is to be upgraded to MLflow 3.

### What needs to get done

Upgrade the charm to MLflow 3, while running upstream images.

### Definition of Done

The charm is upgraded to MLflow 3 and upstream images are employed.

### LINKED ISSUE #433: MLflow 3 Upgrade Testing
### Context

The charm is to be upgraded to MLflow 3.

### What needs to get done

Test the charm upgrade from MLflow 2 to MLflow 3.

### Definition of Done

The charm upgrade from MLflow 2 to MLflow 3 is tested.

MattiaSarti:
resolves https://github.com/canonical/mlflow-operator/issues/425 & resolves https://github.com/canonical/mlflow-operator/issues/433

**note: database migrations are now automatically run on charm upgrades**

**note: many code parts are introduced only to temporarily copy with [an open upstream issue](https://github.com/mlflow/mlflow/issues/19943) and are properly marked as such**


src/charm.py
```diff
@@ -85,6 +84,51 @@
...
+from mlflow.store.db.utils import _verify_schema, create_sqlalchemy_engine_with_retry
+
+engine = create_sqlalchemy_engine_with_retry(sys.argv[1])
+try:
+    _verify_schema(engine)
+except Exception as exc:
+    message = str(exc).lower()
+    if 'Detected out-of-date database schema' in message:
+        print('{SCHEMA_OUT_OF_DATE_MARKER}')
+        sys.exit(0)
```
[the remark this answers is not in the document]
MattiaSarti: Right! It's a typo of mine with the last commits when I was trying to make it more maintainable, but I've fixed it now

src/charm.py
```diff
@@ -784,23 +974,26 @@ def _generate_environment(self) -> dict:
...
         environment_variables = {
-            "MLFLOW_BACKEND_STORE_URI": f"mysql+pymysql://{relational_db_data['username']}:{relational_db_data['password']}@{relational_db_data['host']}:{relational_db_data['port']}/{self._database_name}",  # noqa: E501
+            "MLFLOW_BACKEND_STORE_URI": backend_store_uri,
             "MLFLOW_EXPOSE_PROMETHEUS": METRICS_PATH,
             "MLFLOW_HOST": "0.0.0.0",
             "MLFLOW_PORT": self._mlflow_port,
+            # NOTE: security middleware disable as already provided by the outer Istio layer:
+            # https://mlflow.org/docs/latest/self-hosting/security/network/#disable-security-middleware  # noqa: E501
+            "MLFLOW_SERVER_DISABLE_SECURITY_MIDDLEWARE": "true",
         }
```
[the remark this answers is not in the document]
MattiaSarti: No, because Charmed MLflow without any Istio is not functional and never supposed to be deployed like

src/charm.py
```diff
@@ -85,6 +84,51 @@
 # store's TLS certificate - NOTE: under Pebble's home directory, writable by the non-root user:
 S3_CA_BUNDLE_CONTAINER_PATH = "/var/lib/pebble/default/s3-ca-bundle.pem"
 
+SCHEMA_OUT_OF_DATE_MARKER = "MLFLOW_SCHEMA_OUT_OF_DATE"
+# read-only snippet run in the workload container to detect whether the tracking database schema is
+# outdated with respect to the deployed MLflow version, based on this exception being raised by
+# running `_verify_schema` directly:
+# https://github.com/mlflow/mlflow/blob/v3.14.0/mlflow/store/db/utils.py#L127-L134
```
NohaIhab, line 91: nit: can we put the permalink instead? this one is subject to changes

src/charm.py
NohaIhab, line 1: question: is updating the documentation going to be addressed in a different PR?

.github/automatic_backport_tracks.yaml
```diff
@@ -1,2 +1,2 @@
 automatic_backport_tracks:
-  -   track/2.22
```
NohaIhab, line 2: question: (I could be missing context) Are we going to stop maintaining `2.22`? If not, we can keep both tracks in the backport list.. I believe a portion of the source code is still shared

src/charm.py
```diff
@@ -466,7 +517,146 @@ def _get_relational_db_data(self) -> dict:
...
+            proc = self.container.exec(["python3", "-c", SCHEMA_CHECK_SNIPPET, backend_store_uri])
+            stdout, _ = proc.wait_output()
+
+        except ExecError as error:
+            self.logger.warning(
+                f"Could not determine the database schema state: exit code {error.exit_code}",
+            )
+            raise ErrorWithStatus(
+                "Could not verify the database schema state; will retry.", WaitingStatus
+            )
```
NohaIhab, line 561: nit: can we clarify what it's waiting for here? i.e. db to become reachable

src/charm.py
```diff
@@ -466,7 +517,146 @@ def _get_relational_db_data(self) -> dict:
...
+            self.logger.error(f"Database schema migration failed: exit code {error.exit_code}")
+            raise ErrorWithStatus(
+                "Database schema migration failed. Check the unit logs for details.",
+                # NOTE: Blocked (not Waiting) so the caller does not defer/retry: the container and
+                # database are already reachable and the schema was determinable, so a failing
+                # `mlflow db upgrade` is a non-transient problem (e.g., bad/partial schema,
+                # incompatible version jump, insufficient privileges) that needs operator
+                # intervention rather than a retry:
+                BlockedStatus,
+            )
```
NohaIhab, line 605: suggestion(blocking): do the logs mention guidance on how to resolve the upgrade? I find `BlockedStatus` too harsh if we do not provide instruction how to unblock. What happens in the scenario that the upgrade fails? Can it safely revert to the previous state?

src/charm.py
```diff
@@ -466,7 +517,146 @@ def _get_relational_db_data(self) -> dict:
...
+            # upgrade from a revision that did not request it, since the data-platform provider only
+            # grants extra roles when the relation (and its user) is first created. Retrying will
+            # never succeed, so surface an actionable Blocked status instead of looping in Waiting.
+            if "1227" in stderr or "SYSTEM_VARIABLES_ADMIN" in stderr:
+                raise ErrorWithStatus(
+                    "Database user lacks the SYSTEM_VARIABLES_ADMIN privilege required to migrate "
+                    "the schema. Remove and re-add the 'relational-db' relation (or grant the "
+                    "'charmed_dba' role to the database user) so the charm can proceed.",
+                    BlockedStatus,
+                )
```
NohaIhab, line 654: suggestion(non-blocking): this is a helpful message but I'm not sure how it will show up in the juju `message` because it's somewhat long, can we log details to the debug logs and make this one shorter saying to check logs?

src/charm.py
```diff
@@ -466,7 +517,146 @@ def _get_relational_db_data(self) -> dict:
...
+        """Automatically migrate the tracking database schema when it is out of date.
+
+        Triggered as part of the reconcile logic, notably on the `upgrade-charm` (Juju refresh)
+        event, so that a charm refresh shipping an MLflow release with schema changes migrates the
+        database without manual intervention. A distinctive Maintenance status is set while the
+        migration runs.
+        """
+        backend_store_uri = self._get_backend_store_uri()
+
+        # TODO: remove once this issue is fixed: https://github.com/mlflow/mlflow/issues/19943
```
NohaIhab, line 670: praise: neat TODOs!

src/charm.py
```diff
@@ -466,7 +517,146 @@ def _get_relational_db_data(self) -> dict:
...
+        self._ensure_trigger_creation_allowed(backend_store_uri)
+
+        if self._is_database_schema_out_of_date(backend_store_uri):
+            self.unit.status = MaintenanceStatus(
+                "Database schema is out of date for the deployed MLflow version; migrating it."
+            )
```
NohaIhab, line 676: question: what is setting the status to `active` once the migration is done?

tests/integration/test_charm_major_upgrade.py
```diff
@@ -0,0 +1,290 @@
...
+            status="blocked",
+            raise_on_blocked=False,
+            raise_on_error=False,
+            timeout=600,
+        )
+        # NOTE: waiting for MySQL to reach a stable idle is not desirable, as on relation removal
+        # MySQL can get stuck flapping "executing" while it repeatedly fails to delete the old
+        # scoped user (logging "Failed to delete instance users") - which is nevertheless not a
+        # problem for MLflow, since the stuck teardown of the old user does not affect re-adding
+        # the relation, which creates a fresh scoped user granted the newly requested privileges
```
NohaIhab: note: this seems like a MySQL bug, have you considered reporting it?

tests/integration/test_charm_major_upgrade.py
```diff
@@ -0,0 +1,290 @@
...
+        with binary logging enabled, MySQL rejects for a user lacking global privileges; the locally
+        built charm requests the `charmed_dba` role via `extra_user_roles` (granting
+        SYSTEM_VARIABLES_ADMIN and TRIGGER) and, using those relation credentials, persists
+        `log_bin_trust_function_creators` before the migration runs; the data-platform provider only
+        grants extra roles on the *first* `database_requested` event, which already fired for the
+        old version without that role, so we remove the `relational-db` relation before refreshing
+        and re-add it afterwards so the request fires again for the new charm; the pre-upgrade data
+        is preserved because dropping the relation only deletes the scoped user, not the database
+        """
+        # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
```
NohaIhab, line 212: question: what's the `- - -` comment style? It's ok, just not consistent with our source(s).

tests/integration/test_charm_major_upgrade.py
NohaIhab, line 1: praise: Really cool tests!! 🤩

src/charm.py
```diff
@@ -841,6 +1034,45 @@ def _remove_authorization_policies(self, _):
...
+
+        except ErrorWithStatus as err:
+            self.model.unit.status = err.status
+            self.logger.info(f"Event {event} stopped early with message: {str(err)}")
+
+            # NOTE: deferring only when waiting, so that a genuine migration failure (Blocked) is
+            # surfaced and not retried while the migration is instead retried when waiting for the
+            # workload to be ready or for something else:
+            if isinstance(err.status, WaitingStatus):
+                event.defer()
```
NohaIhab, line 1094: IMO defer makes sense here, but historically the team is not a fan :D we can bring this to the others

NohaIhab (changes requested):
Thank you @MattiaSarti, nicely done! Left some comments.

tests/integration/test_charm_major_upgrade.py
misohu, line 1: I am little confused why do we need this test ? .... the idea of major version bump is that it is not backwards compatible ... so this should not work ? I see it might work for certain functionality of mlflow maybe the experiment example you do here but for sth i t will not work thats why they bumped the major version right ?

.github/workflows/ci.yaml
```diff
@@ -131,6 +131,7 @@ jobs:
           - integration-object-storage
           - integration-s3
           - integration-ambient
+          - integration-major-upgrade
```
misohu, line 134: Do we want to keep this on every PR tho ? when we now move to other minor bumps this test might break right ? as upstream will not guaranteed the backwards compatibility for version 3 with version 2

MattiaSarti: Yes, it will be possible to upgrade until we stay on 3.X, and this test passing right now proves that from 2.X to 3.14 we can, and from 3.14 to 3.X we will (because the major will still be the same), so we'll have to worry about that only when we migrate to 4.X, if ever.

Also, see these docs that suggest migrating from 2.X to 3.X is possible:
- https://mlflow.org/docs/latest/ml/mlflow-3/#can-mlflow-3x-load-resources-created-with-mlflow-2x
- https://mlflow.org/docs/latest/self-hosting/migration/

tests/integration/test_charm_major_upgrade.py
MattiaSarti: It is compatible, instead, at least for us relying on a remote backend store and on a remote artifact store (if you rely on the local filesystem, it is not) - see these docs that suggest migrating from 2.X to 3.X is possible exactly how this PR implements it in our charm:
- https://mlflow.org/docs/latest/ml/mlflow-3/#can-mlflow-3x-load-resources-created-with-mlflow-2x
- https://mlflow.org/docs/latest/self-hosting/migration/

src/charm.py
MattiaSarti: I put the one of the current version on purpose because, if the source changes with newer versions, then the reference to those lines won't match what excepted anymore

src/charm.py
MattiaSarti: Yes, we have separate tasks for that:
- https://github.com/canonical/mlflow-operator/issues/431
- https://github.com/canonical/mlflow-operator/issues/432

.github/automatic_backport_tracks.yaml
MattiaSarti: Yes, we will keep maintaining both versions, I simply thought that we would raise PRs for 3.X more often and so that it was simpler to manually add labels for 2.X backports rather to manually remove them most of the times, on future PRs, but I can reintroduce it if you think we should

src/charm.py
MattiaSarti: This runs within the event handler for `upgrade-charm` events, and then the general `_on_event` handler is guaranteed to run right after and to set the status to `active` if all goes well, since [on charm upgrades Juju emits other events such as `config-changed` after `upgrade-charm`](https://canonical.com/juju/docs/juju-cli/3.6/reference/hook/#upgrade-charm)

tests/integration/test_charm_major_upgrade.py
MattiaSarti: Yes, I did and I am still waiting for them to reply

tests/integration/test_charm_major_upgrade.py
MattiaSarti: It was just to make it clear that all and only the multi-line section delimited by those characters was to be removed when the bug was fixed:

image

Is there some more consistent way that we could signal that?

src/charm.py
MattiaSarti: Sure! Let's hear @misohu's opinion

src/charm.py
MattiaSarti: (the updated logs explain that all)

src/charm.py
misohu: hmm I understand that this is only executed during update charm hook so we dont have standard defer problem here .... there is also no other way how to solve it

tests/integration/test_charm_major_upgrade.py
misohu, line 1: I would add a test after the upgrade that we can log something to new version of mlflow .... just to test that its working also there

.github/automatic_backport_tracks.yaml
NohaIhab: My preference is to have both and decide to remove it per PR, otherwise we might forget about the other track  IMO.

src/charm.py
NohaIhab: SGTM, resolving

tests/integration/test_charm_major_upgrade.py
NohaIhab: ah okay it's clear to me now

tests/integration/test_charm_major_upgrade.py
NohaIhab: we should link the issue in the comment then

NohaIhab (commented):
Thanks Mattia, just a couple things pending

tests/integration/test_charm_major_upgrade.py
NohaIhab: since the issue is not coming up anymore I'm resolving this

NohaIhab (approved):
Thanks @MattiaSarti 🚀

misohu (approved):

MattiaSarti:
note: force-pushing to rebase before merging

Issue #456, opened 2026-08-17
Custom Authentication Logic for Multi-Tenancy


MattiaSarti:
### Context

The charm has to support multi-tenancy according to the design outlined in the official spec.

### What needs to get done

Add custom authentication logic to MLflow so that MLflow's internal RBAC is mapped to identities of the identity stack and of Kubeflow Profiles.

### Definition of Done

Custom authentication plugin is added to MLflow so that MLflow's internal RBAC is mapped to identities of the identity stack and of Kubeflow Profiles.

MattiaSarti:
Internal ticket manually created, because of a bug with Jira's automation (reason why the label for Jira was removed): https://warthogs.atlassian.net/browse/KF-8905

PR #485, opened 2026-09-02
feat: add custom authentication for multi-tenancy

MattiaSarti:
resolves https://github.com/canonical/mlflow-operator/issues/456

note: since problems with MySQL for RBAC similar to https://github.com/mlflow/mlflow/issues/9642 were encountered, it was necessary to also migrate from MySQL to PostgreSQL for RBAC to be functional — a task that was already part of our roadmap anyway — and as a nice side effect the workaround introduced in the past for database trigger creation could be removed


metadata.yaml
```diff
@@ -27,7 +27,7 @@ resources:
     upstream-source: docker.io/charmedkubeflow/mlflow-prometheus-exporter:3.15.1-340fe92-20260826095012
 requires:
   relational-db:
-    interface: mysql_client
+    interface: postgresql_client
```
deusebio, line 30: **note** I understand that the [the bug](https://github.com/mlflow/mlflow/issues/9642) is effectively preventing us to use MySQL in the first place, but this will mean that we are going to ONLY support postgresql. I would probably try to be more general and rename the "relational-db" into "postgresql" such that - if the bug will ever get resolved, we can also re-introduce MySQL.

Let's also align @grobbie about this, since this will likely means that having a Kubeflow+MLFlow deployment only backed by MySQL database is not going to be possible with MLFlow 3

src/auth/custom_userid_header_auth.py
```diff
@@ -0,0 +1,98 @@
...
+    """Just-in-time provision a dormant, grant-less MLflow user for a newly seen identity.
+
+    The user is created with no grants, so deny-by-default RBAC denies it every operation until it
+    is granted access. Its password is random and never used: this user authenticates through the
+    trusted identity header, not Basic auth. Provisioning grants presence, never access.
+    """
+    if store.has_user(username):
+        return
+    try:
+        store.create_user(username, secrets.token_urlsafe(32), is_admin=False)
```
deusebio, line 50: **question** this is quite strange to see. Is this the MLFlow guidance? I'm just afraid that this is basically allowing non-existing users to authenticate. Sure, with a 0 permissions role, but if - by chance - an endpoint would be open to all authenticated users (not having authz rules) that could maybe be exploited.

But anyhow, I would have prefer if we could have return a Unauthorized code straight away, but if this guidance is provided by MLFlow, I suppose that's fine.

src/charm.py
```diff
@@ -1039,6 +1012,55 @@ def _generate_environment(self) -> dict:
...
+        except SecretNotFoundError:
+            if not self.unit.is_leader():
+                raise ErrorWithStatus(
+                    "Waiting for the leader to generate the RBAC credentials", WaitingStatus
+                )
+            content = {
+                "flask-secret-key": secrets.token_urlsafe(32),
+                "admin-password": secrets.token_urlsafe(32),
+            }
+            self.app.add_secret(content, label=AUTH_SECRET_LABEL)
```
deusebio, line 1037: **todo** I suppose that if we will enable HA, the secret will need to be stored in a peer relations such that the admin credentials can be shared among the various units. Possibly out of scope in this PR if the HA will be targeted later, but it is probably something worth thinking of already

src/charm.py
```diff
@@ -1102,9 +1124,11 @@ def _on_pebble_ready(self, _):
...
         self._on_event(_)
 
-    def _on_database_relation_removed(self, _) -> None:
-        """Event is fired when relation with postgres is broken."""
-        self.unit.status = BlockedStatus("Please add relation to the database")
+    def _on_backend_store_relation_removed(self, _) -> None:
+        """Event is fired when relation with the backend store is broken."""
+        self.unit.status = BlockedStatus(
+            f"Please add the relation {RELATION_ENDPOINT_FOR_BACKEND_STORE_DB}"
+        )
```
deusebio, line 1141: **question** I see that it was not done even before, but when we remove this, beside blocking, shouldn't we also stop/reconfigure the service? Otherwise the service will keep on running, still pointing a the db, no?

tests/integration/test_charm_object_storage.py
```diff
@@ -372,14 +379,65 @@ async def test_can_create_experiment_with_mlflow_library(self, ops_test: OpsTest
...
         assert response.status_code == 200
         client.create_experiment(TEST_EXPERIMENT_NAME)
         all_experiments = client.search_experiments()
         assert len(list(filter(lambda e: e.name == TEST_EXPERIMENT_NAME, all_experiments))) == 1
 
         mlflow_subprocess.terminate()
 
+    # TODO: update this test's logic as multi-tenancy is developed:
+    @pytest.mark.abort_on_fail
+    async def test_mlflow_user_identity_and_tenant_rbac(self, ops_test: OpsTest):
```
deusebio: **question** is this repeated in the various tests? If so, wouldn't it be worth making it parametric and abstracting that away in a single functions in the helpers?

deusebio (commented):
The changes looks mostly sane! I only have a few comments, mostly due to my limited understanding of the charm logic here. 

I believe the changes to swap to postgresql are sane given the existing bug, I would just maybe consider to ALSO support mysql that would allow us to have a smaller deployment (which is something Rob has recently been raising). So to do so, I would probably rename the endpoint such that we could eventually also add MySQL, should the bug ever get solved (let's nevertheless align him tomorrow at the daily)

tests/integration/test_charm_object_storage.py
MattiaSarti: Yes, it is. I would, but the problem is that we don't do this just for this test case, we do this for all tests and across all charms! When migrating all charms to ambient mode, there were some discussions with Kimonas and it was decided to intentionally keep the same exact test cases defined (hence repeated) twice, across both ambient-mode and non-ambient-mode test files, so that we didn't have to waste time refactoring all tests and in the future we could just get rid of the non-ambient-mode files when only ambient mode would be supported. Then, when migrating to S3 from object storage, the same pattern was followed. As a result, we now repeat all test cases across all test files in all of our charms, knowing we'll simply eventually remove object-storage-based, non-ambient-mode test files when only ambient mode and S3 storage will be supported. So I can parametrize this test case here, but it would solve 0.1% of the repetition problem and it would introduce kind of an anti-pattern. Shall we leave it this way, then?

src/charm.py
MattiaSarti: (By "service" you mean the tracking server workload, right?)
Yes, it's a fair point. I can introduce this mechanism, if we want.
Otherwise, the server will reject requests until its persistence layer becomes available again.

src/charm.py
MattiaSarti: Actually, I initially went for peer relations to then realize secrets were the proper way of handling this. To the best of my knowledge, peer relations and secrets are _**complementary**_ mechanisms. This already scales to multiple units, for when we'll enable them targeting HA, and the secrets is already encrypted and stored by Juju, and as you can see in this source only the leader writes it and non-leaders read it, so this already allows sharing information across units securely and durably. Peer relations would be an alternative way of persisting information and sharing it across units when it's non-sensitive, because it is stored in relation databags and not encrypted.

metadata.yaml
MattiaSarti: Correct. Sure! I'm open to any suggestions

src/auth/custom_userid_header_auth.py
MattiaSarti: It's exactly as you said, with one important caveat: by design, it will be impossible to reach the tracking server without a counterfeit user-ID header and therefore to compromise authentication and to exploit the system. This will be guaranteed by the mesh around the charm. And we'll always have RBAC on top of this, with zero grants be default. So authorization will be effective and there would be no way a newly seen user could do anything unless explicitly configured via `data-integrator` or `data-kubeflow-integrator`. What's more, this just-in-time user provisioning is necessary and also crucial to the user experience, as outlined in our spec.

deusebio: Oh I see. I believe it would be nice to have a config option to create users on access, such that we can tune this behaviour

tests/integration/test_charm_object_storage.py
deusebio: That's fine. If that's the pattern currently implemented, it is not for this PR to change this.

src/charm.py
deusebio: All right! Check with Neha about the grants of the secrets! Thanks!

tests/integration/test_charm_ambient.py
NohaIhab, line 1: suggestion(blocking): Include a test for the access denied case, for example with the incorrect Identity, to make sure the server is really protected

src/charm.py
NohaIhab, line 1: suggestion(blocking): let's also update the `releases/` with postgresql, for the `latest/edge` channel

Additionally, the docs would need to be updated, but this should be aligned with once we release the new version, can you create a ticket to track that?

src/auth/custom_userid_header_auth.py
```diff
@@ -0,0 +1,98 @@
...
+        return authenticate_request_basic_auth()
+
+    # when neither authentication method is provided, deny because unauthenticated - inspired by:
+    # https://github.com/mlflow/mlflow/blob/v3.15.1/mlflow/server/auth/__init__.py#L494-L502
+    missing_authentication_resonse = make_response(
+        "Not authenticated. Pass either the expected identity header or Basic auth credentials."
+    )
+    missing_authentication_resonse.status_code = 401
+    missing_authentication_resonse.headers["WWW-Authenticate"] = 'Basic realm="mlflow"'
+    return missing_authentication_resonse
```
NohaIhab: nitpick: Typo `missing_authentication_resonse` -> `_response`

tests/unit/test_operator.py
NohaIhab, line 1: suggestion (blocking): Add unit tests for `_get_or_create_auth_secrets` and `reconcile_auth_config` . These are the core of the new feature and are only exercised via patched mocks.

tests/integration/test_charm_upgrade_minor.py
```diff
@@ -111,10 +107,15 @@ def __exit__(self, *exc):
...
-    response = requests.get(url)
+    response = requests.get(
+        url,
+        # TODO: remove once multi-tenancy is completed:
+        headers={IDENTITY_HEADER_NAME: TEST_IDENTITY},
+    )
     assert response.status_code == 200
 
 
+@pytest.mark.skip(reason="TODO: restore once we have something on stable for MLflow 3")
```
NohaIhab, line 118: suggestion: create an issue to track re-enabling upgrade tests

src/auth/custom_userid_header_auth.py
NohaIhab: this is a good idea, so operators can opt into a stricter "pre-provisioned users only" mode

NohaIhab (changes requested):
thanks @MattiaSarti, reviewed

src/auth/custom_userid_header_auth.py
MattiaSarti: As agreed via call, tracked with https://github.com/canonical/mlflow-operator/issues/488

tests/integration/test_charm_ambient.py
MattiaSarti: Done! (The test actually asserts that access is not denied but grant-less, as we have just-in-time user creation.)

tests/integration/test_charm_upgrade_minor.py
MattiaSarti: Nice idea, [here](https://github.com/canonical/mlflow-operator/issues/489) it is!

src/charm.py
MattiaSarti: Done! But for the docs we'll have some huge work to be done for the epic, not just this change, and we'll create all docs' tasks at a later point. Do you find that reasonable?

metadata.yaml
MattiaSarti: Agreed via call that we can get rid of MySQL right now without ever reintroducing it.

src/charm.py
MattiaSarti: Implemented!

src/charm.py
NohaIhab: yes that's ok

NohaIhab (approved):
thank you @MattiaSarti

deusebio (approved):
Looks good!

Issue #475, opened 2026-08-30
Removal of Charm Config for NodePort as K8s Service Kind


MattiaSarti:
### Context

The charm has to support multi-tenancy according to the design outlined in the official spec.

### What needs to get done

Remove the charm config that allows to parametrize the kind of the K8s Service to the tracking server as either ClusterIP or NodePort, so that it is always and only ClusterIP.

### Definition of Done

The charm config that allows to parametrize the kind of the K8s Service to the tracking server as either ClusterIP or NodePort is removed, so that it is always and only ClusterIP.

MattiaSarti:
Internal ticket manually created, because of a bug with Jira's automation (reason why the label for Jira was removed): https://warthogs.atlassian.net/browse/KF-8941

PR #486, opened 2026-09-07
feat: remove NodePort config

MattiaSarti:
resolves https://github.com/canonical/mlflow-operator/issues/475

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -12,25 +12,11 @@ options:
       The port MLFlow will be listening on
     type: int
     default: 5000
-  mlflow_nodeport:
-    description: |
-      The nodeport for MLFlow
-    type: int
-    default: 31380
   mlflow_prometheus_exporter_port:
     description: |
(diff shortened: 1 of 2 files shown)


config.yaml
```diff
@@ -12,25 +12,11 @@ options:
       The port MLFlow will be listening on
     type: int
     default: 5000
-  mlflow_nodeport:
```
dariofaccin, line 15: thought: since we are removing a config option, we should be particularly careful as it might break backward compatibility and prevent smooth refreshes.

MattiaSarti: Absolutely right! But we are intentionally cumulating all non-backward-compatible changes now that we haven't released to stable yet for MLflow 3 (and multi-tenancy will bring many such changes), so this backward incompatibility is already accounted for

mvlassis (approved):

dariofaccin (approved):

Issue #474, opened 2026-08-30
Charm Config for Identity Aliases for Multi-Tenancy


MattiaSarti:
### Context

The charm has to support multi-tenancy according to the design outlined in the official spec.

### What needs to get done

Add a new charm config for identity aliases so that the tracking server's custom authentication logic can map different user-ID headers to the same MLflow user as aliases in such a charm config.

### Definition of Done

A new charm config for identity aliases is added so that the tracking server's custom authentication logic can map different user-ID headers to the same MLflow user as aliases in such a charm config.

PR #487, opened 2026-09-07
feat: allow user aliasing

MattiaSarti:
resolves https://github.com/canonical/mlflow-operator/issues/474

### DIFF
--- a/src/auth/custom_userid_header_auth.py
+++ b/src/auth/custom_userid_header_auth.py
@@ -36,7 +36,7 @@ HARDCODED_TEST_WORKSPACE = "default"
 # NOTE: there environment variables correspond to charm configs and are guaranteed to always be up
 # to date because the charm reloads the tracking server on config changes:
 _IDENTITY_HEADER_NAME = os.environ["IDENTITY_HEADER_NAME"]
-_RAW_IDENTITY_ALIASES = os.environ.get("IDENTITY_ALIASES")
+_RAW_IDENTITY_ALIASES = os.environ["IDENTITY_ALIASES"]
 _IDENTITY_ALIASES = json.loads(_RAW_IDENTITY_ALIASES) if _RAW_IDENTITY_ALIASES else {}
 
 
--- a/src/auth/custom_userid_header_auth.py
+++ b/src/auth/custom_userid_header_auth.py
@@ -33,7 +33,7 @@ from werkzeug.datastructures import Authorization
 HARDCODED_TEST_IDENTITY = "charm-test-user"
 HARDCODED_TEST_WORKSPACE = "default"
 
-# NOTE: there environment variables correspond to charm configs and are guaranteed to always be up
+# NOTE: these environment variables correspond to charm configs and are guaranteed to always be up
 # to date because the charm reloads the tracking server on config changes:
 _IDENTITY_HEADER_NAME = os.environ["IDENTITY_HEADER_NAME"]
 _RAW_IDENTITY_ALIASES = os.environ["IDENTITY_ALIASES"]
--- a/src/auth/custom_userid_header_auth.py
+++ b/src/auth/custom_userid_header_auth.py
@@ -30,10 +30,12 @@ from mlflow import MlflowException
 from mlflow.server.auth import authenticate_request_basic_auth, store
 from werkzeug.datastructures import Authorization
 
+
 class IdentityAliasesCache(TypedDict):
     path: str
     modified_time: float
     aliases: dict[str, str]
+
 
 # in-process cache of the charm-maintained identity-aliases file, reloaded only when the file's
 # modification time changes so per-request cost stays negligible while alias edits still take
--- a/src/auth/custom_userid_header_auth.py
+++ b/src/auth/custom_userid_header_auth.py
@@ -29,11 +29,9 @@ from mlflow import MlflowException
 from mlflow.server.auth import authenticate_request_basic_auth, store
 from werkzeug.datastructures import Authorization
 
-
 # TODO: remove once multi-tenancy is completed:
 HARDCODED_TEST_IDENTITY = "charm-test-user"
 HARDCODED_TEST_WORKSPACE = "default"
-
 
 # NOTE: there environment variables correspond to charm configs and are guaranteed to always be up
 # to date because the charm reloads the tracking server on config changes:
--- a/src/auth/custom_userid_header_auth.py
+++ b/src/auth/custom_userid_header_auth.py
@@ -10,8 +10,8 @@ stack and service mesh) sets securely, and falls back to HTTP Basic auth only fo
 callers reaching the server over localhost (the charm super-admin and the metrics exporter).
 
 The header name is read from the ``IDENTITY_HEADER_NAME`` environment variable, and the optional
-identity-alias map from the file at ``IDENTITY_ALIASES_PATH``, both set by the charm from its
-``identity_header_name`` and ``identity_aliases`` config options, respectively.
+identity-alias map from the ``IDENTITY_ALIASES`` environment variable, both set by the charm from
+its ``identity_header_name`` and ``identity_aliases`` config options, respectively.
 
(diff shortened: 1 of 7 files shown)


src/auth/custom_userid_header_auth.py
```diff
@@ -9,24 +9,43 @@
...
+
+
+# in-process cache of the charm-maintained identity-aliases file, reloaded only when the file's
+# modification time changes so per-request cost stays negligible while alias edits still take
+# effect without a server restart:
+_IDENTITY_ALIASES_CACHE: IdentityAliasesCache = {
+    "file_path": "",
+    "file_modified_time": 0.0,
+    "aliases": {},
+}
```
deusebio: **thoughts** not against, but this seems a bit over-engineered in the sense that I don't think this file can ever change, unless the charm is doing that, and at that point, when someone sets a new config options we are likely to restart the service anyway. So even just reading the file would probably do to keep the LoC down. It is not a big point, it just makes the code a bit more complex for no reason, but up to you.

src/auth/custom_userid_header_auth.py
```diff
@@ -37,6 +56,40 @@ def _identity_header_name() -> str:
...
+
+    try:
+        identity_aliases_file_modified_time = os.path.getmtime(identity_aliases_file_path)
+    except OSError:
+        return {}  # no identity aliases configured
+
+    # updating the cache if the file defining identity aliases has changed, in particular...
+    if (
+        # ... if the location of the file has changed:
+        _IDENTITY_ALIASES_CACHE.get("file_path") != identity_aliases_file_path
```
NohaIhab: question: why would the file location change? I see it's a constant in charm.py

src/auth/custom_userid_header_auth.py
NohaIhab: issue(blocking): The `TypedDict` declares `path`/`modified_time`, but the literal and all subsequent accesses use `file_path`/`file_modified_time`.  We did not catch this in CI because we don't run static type checkers.. maybe it's something we should consider.

src/auth/custom_userid_header_auth.py
```diff
@@ -37,6 +56,40 @@ def _identity_header_name() -> str:
     return os.environ["IDENTITY_HEADER_NAME"]
 
 
+def _identity_aliases() -> dict:
```
NohaIhab: suggestion(blocking): I think the return type here should be `dict[str, str]`

src/auth/custom_userid_header_auth.py
```diff
@@ -76,11 +129,15 @@ def authenticate_request() -> Authorization | Response:
     """
     identity = request.headers.get(_identity_header_name())
     if identity:
-        # ensuring that the MLflow user whose username matches the identity header's value either
-        # already exists or is successfully created, for downstream permission resolution to work:
-        _ensure_user_exists(identity)
+        # resolving the identity to its primary identity, which is the MLflow username: the
+        # identity is the username unless the charm-maintained alias file remaps a secondary
+        # identity onto a primary one:
+        username = _identity_aliases().get(identity, identity)
```
NohaIhab: nit: this comment is a bit confusing, I prefer if it's written to explain the different cases and what the output is for each case

tests/integration/test_charm_ambient.py
NohaIhab, line 1: suggestion(non-blocking): can we add some integration tests covering the modification-time-based cache reload? for example, that the alias change takes effect live, or that clearing aliases reverts resolution.

tests/unit/test_operator.py
NohaIhab, line 1: suggestion(non-blocking):  can we add a test to check that `_identity_aliases` returns the file's map, reloading only when the file's path or mod. time changes? we can cover the different cases in a parametrized test.

src/charm.py
```diff
@@ -1029,6 +1036,59 @@ def _get_or_create_auth_secrets(self) -> dict:
...
+                "'<secondary-identity>: <primary-identity>' entries.",
+                BlockedStatus,
+            )
+
+        for identity in (*aliases, *aliases.values()):
+            if not isinstance(identity, str) or not identity.strip():
+                raise ErrorWithStatus(
+                    "Config 'identity_aliases' identities must be non-empty strings; quote any "
+                    "value that YAML would otherwise read as a number or boolean.",
+                    BlockedStatus,
```
NohaIhab: nit: message might be too long for the status, if we can make it a bit shorter, or mention to check the debug log and log the error.. this might apply to other messages too, but I'll only leave the one comment here.

src/auth/custom_userid_header_auth.py
```diff
@@ -37,6 +56,40 @@ def _identity_header_name() -> str:
...
+    # updating the cache if the file defining identity aliases has changed, in particular...
+    if (
+        # ... if the location of the file has changed:
+        _IDENTITY_ALIASES_CACHE.get("file_path") != identity_aliases_file_path
+        # ... or if the modification time of the file has changed:
+        or _IDENTITY_ALIASES_CACHE.get("file_modified_time") != identity_aliases_file_modified_time
+    ):
+        # reloading the aliases from the updated file:
+        with open(identity_aliases_file_path, encoding="utf-8") as identity_aliases_file:
+            _IDENTITY_ALIASES_CACHE["aliases"] = json.load(identity_aliases_file)
```
NohaIhab: suggestion(non-blocking): add schema validation for the JSON

src/auth/custom_userid_header_auth.py
```diff
@@ -37,6 +56,40 @@ def _identity_header_name() -> str:
     return os.environ["IDENTITY_HEADER_NAME"]
 
 
+def _identity_aliases() -> dict:
+    """Return the charm-maintained flat ``{secondary-identity: primary-identity}`` alias map.
+
+    The map is read from the file whose path the charm sets in ``IDENTITY_ALIASES_PATH`` and cached
+    in-process, reloaded only when the file changes, so alias edits take effect without a restart.
+    An absent path or file (or an empty map) means no aliases are configured.
```
NohaIhab: thought: while I like the smooth reload feature, I'm a little scared of how this plays out in the charm, in the past we've handled this with a pebble push and replan/restart. We can discuss about this one.

NohaIhab (changes requested):
thanks @MattiaSarti, reviewed

MattiaSarti:
> thought: while I like the smooth reload feature, I'm a little scared of how this plays out in the charm, in the past we've handled this with a pebble push and replan/restart. We can discuss about this one.

It's a valid concern, @NohaIhab, and we can get rid of the file-caching mechanism and simply reload the tracking server on config changes, even if not super-efficient, because config changes are sporadic events and we can afford losing some availability on them, if that helps avoid bugs and keep the code more maintainable.

src/auth/custom_userid_header_auth.py
MattiaSarti: → https://github.com/canonical/mlflow-operator/pull/487#issuecomment-5635330158

src/auth/custom_userid_header_auth.py
MattiaSarti: → https://github.com/canonical/mlflow-operator/pull/487#issuecomment-5635330158

tests/unit/test_operator.py
MattiaSarti: → https://github.com/canonical/mlflow-operator/pull/487#issuecomment-5635330158

tests/integration/test_charm_ambient.py
MattiaSarti: → https://github.com/canonical/mlflow-operator/pull/487#issuecomment-5635330158

deusebio (approved):
I had already started the review on Friday, and as a matter of fact I had the very same comment as @NohaIhab (te comment still refers to the outdated version of the file), which has already been addressed in the review.

So I don't have futher concerns and the PR is good to go

NohaIhab (approved):
Thank you!

Issue #492, opened 2026-09-11
`mlflow_client` Provider for Tenant Provisioning and User Granting


MattiaSarti:
### Context

The charm has to support multi-tenancy according to the design outlined in the official spec.

### What needs to get done

Implement the `mlflow_client` logic for tenant provisioning and user granting on the MLflow charm's side, that is the provider.

### Definition of Done

The `mlflow_client` logic for tenant provisioning and user granting is implemented on the MLflow charm's side, that is the provider.

PR #494, opened 2026-09-15
feat: add mlflow_client provider logic

MattiaSarti:
resolves https://github.com/canonical/mlflow-operator/issues/492


MattiaSarti:
note: integration tests were added in the commit right before this comment, but they are expected to fail until the current data-integrator changes for the requirer's logic of `mlflow_client` don't land on `latest/edge`, since there's no charm that can be deployed for that new version of data-integrator until its changes live only on the branch of a pull request - see [this comment](https://github.com/canonical/data-integrator/pull/328#discussion_r4050647700) for the rationale

src/charm.py
```diff
@@ -1157,6 +1183,178 @@ def _reconcile_auth_config(self) -> None:
...
+                        relation.id,
+                        ResourceProviderModel(
+                            request_id=request.request_id,
+                            # NOTE: while `entity_name` in the provider's response is not used to
+                            # provide clients with that information, as clients themselves set it
+                            # for the provider on the requirer's side in the first place and can
+                            # therefore already be informed about its value by the requirer,
+                            # including it here, in the provider's response, is nevertheless
+                            # necessary for the requirer to correlate the response with its request
+                            # and for the functionality of the employed library:
```
NohaIhab: nit: this is a helpful comment but can you make it less verbose

tests/integration/test_charm_ambient.py
```diff
@@ -318,61 +359,271 @@ async def test_logging(self, ops_test: OpsTest):
...
+            found = minio_client.bucket_exists(bucket_name=default_bucket_name)
+            assert found, f"The '{default_bucket_name}' bucket does not exist"
+
+    @pytest.mark.abort_on_fail
+    async def test_deploy_data_integrator(self, ops_test: OpsTest):
+        """Deploy a data-integrator instance, for user grants in subsequent tests."""
+        # TODO: remove this command and restore the command below once
+        # https://github.com/canonical/data-integrator/pull/328 lands on main, that is on channel
+        # "latest/edge", and mind that an explicit Juju-CLI deploy is temporarily required because
+        # python-libjuju's `Model.deploy()` breaks with this temporary channel format:
```
NohaIhab: todo: remove this now that #328 is merged

src/mlflow_client_reconcile.py
```diff
@@ -0,0 +1,158 @@
...
+import json
+import secrets
+import sys
+
+import sqlalchemy
+from mlflow import MlflowException
+from mlflow.entities.workspace import Workspace
+from mlflow.server.auth import store
+from mlflow.server.auth.config import read_auth_config
+from mlflow.server.workspace_helpers import _get_workspace_store
```
NohaIhab: suggestion(blocking): we are using internal APIs here, a future MLflow version could change those, which would be a maintenance cost for us. Could we instead use [the Authentication REST API](https://mlflow.org/docs/latest/api_reference/auth/rest-api.html#mlflow-authentication-rest-api)? That could be a bit safer than direct store access..

src/mlflow_client_reconcile.py
```diff
@@ -0,0 +1,158 @@
...
+    for workspace, tier in entry["workspace_grants"]:
+        _tolerate_already_exists(
+            workspace_store.create_workspace, Workspace(name=workspace, description=None)
+        )
+        role = _get_or_create_role(role_name, workspace)
+        native_grants = TIERS_TO_NATIVE_GRANTS.get(tier, TIERS_TO_NATIVE_GRANTS[DEFAULT_TIER])
+        for resource_type, resource_pattern, permission in native_grants:
+            _tolerate_already_exists(
+                store.add_role_permission, role.id, resource_type, resource_pattern, permission
+            )
```
NohaIhab: question(blocking): what if a user's tier changes? we should then remove stale permissions from an existing role. If a role exists, `_get_or_create_role` returns it unchanged right? It's not checking if permissions changed. I could be missing something here 🤔

metadata.yaml
```diff
@@ -136,6 +136,14 @@ requires:
     description: |
       Subscribe this charm into a service mesh to enforce authorization policies.
 provides:
+  mlflow-client:
+    interface: mlflow_client
+    description: |
+      Provisions, for a requirer such as data-integrator, an MLflow user the given username and the
+      given workspace grants, and returns no information, since the username is same the as
```
NohaIhab: > and returns no information

question: doesn't it return the entity name?

tests/unit/test_operator.py
```diff
@@ -2171,3 +2173,337 @@ def test_ambient_mode_ingress_listener_port(
...
+    def test_relation_broken_reconciles_excluding_relation(self, harness: Harness):
+        harness.begin()
+        relation_id, _ = add_relation(harness, MLFLOW_CLIENT_RELATION_ENDPOINT)
+
+        reconcile_mock = MagicMock()
+        harness.charm._reconcile_mlflow_client_grants = reconcile_mock
+
+        harness.remove_relation(relation_id)
+
+        reconcile_mock.assert_called_once_with(exclude_relation_id=relation_id)
```
NohaIhab: suggestion(non-blocking): also verify the resulting roles or permissions after removal

tests/integration/test_charm_ambient.py
```diff
@@ -96,6 +109,34 @@
...
+            [
+                "kubectl",
+                "-n",
+                self._namespace,
+                "port-forward",
+                f"svc/{self._charm_name}",
+                f"{self._port}:{self._port}",
+            ]
+        )
+        time.sleep(10)  # waiting for the port-forwarding to be established
```
NohaIhab: suggestion(non-blocking): can we have a proper wait for this one? for example by trying to reach the service?

tests/integration/test_charm_ambient.py
NohaIhab: suggestion(non-blocking): applies to multiple tests, can we assert exact expected role names/counts and permission sets, rather than only validating that roles exist? for example in `test_configured_workspace_grants_are_defined`

NohaIhab (commented):
Thanks @MattiaSarti, great effort, did a first pass

tests/integration/test_charm_ambient.py
MattiaSarti: Unfortunately, even if merged, it did not land on `latest/edge` yet - see [their red CI on main](https://github.com/canonical/data-integrator) and [this Mattermost post about their problems](https://chat.canonical.com/canonical/pl/o3jo8tq6it8m58w5zmw1upin1y) - so that will have to be restored in a subsequent PR

src/mlflow_client_reconcile.py
MattiaSarti: I see your point, but as commented below `def _exec_mlflow_client_reconcile` in `src/charm.py`, 
we need that now that we allow granting super-admins:
> Run a Python script in the tracking server's container to reconcile users' grants across
        workspaces as per the given, desired set of grants. The script acts directly on MLflow's
        auth store instead of calling the tracking server's HTTP API, so that the reconcile never
        has to authenticate to MLflow and keeps working even if the charm's super-admin was changed
        or deleted (intentionally or inadvertently) by an external super-admin.

I can add some docstring comment to the script with our current version number so that it's easier for us to remember to check that when searching version matched when upgrading the charm. Would that help?

metadata.yaml
MattiaSarti: Technically speaking yes, but it's just to make the library's behavior not break, while in reality it does not return meaningful information (the requirer already has that information, together with other essential ones, and we don't add anything from the provider). Shall I specify that anyway, or would it create confusion?

src/mlflow_client_reconcile.py
MattiaSarti: No, you are absolutely right! This is something I missed, let me fix that.

tests/unit/test_operator.py
MattiaSarti: We do have that but in integration tests: `test_removing_relation_revokes_workspace_grants`. Because it's easier to test there that in unit tests, being that outcome due to the reconcile script and not the charm source. Here, in unit tests, we just ensure that the reconcile script is called. Then integration tests cover all effects of that.

src/mlflow_client_reconcile.py
MattiaSarti: I've also added integration tests to cover this, because I only had them change grants across workspaces, not within the same workspace

tests/integration/test_charm_ambient.py
MattiaSarti: Sure. `test_configured_workspace_grants_are_defined` and similar tests already verify role permissions across different workspaces, though, covering the role recreation on workspace changes. I've instead added separate tests to make sure that also role updates are covered, without recreating roles because workspaces change but ensuring the same roles' old permissions are removed, as per https://github.com/canonical/mlflow-operator/pull/494#discussion_r4091626403.

src/mlflow_client_reconcile.py
NohaIhab: do you think it's too complex or it doesn't make sense to limit the exposure to MLflow internals to the one case that actually needs it? i.e. when reconciling a super-admin grant

src/mlflow_client_reconcile.py
NohaIhab: thank you

metadata.yaml
NohaIhab: ok, no I wouldn't overload the metadata.yaml, we can add a comment when defining the provider

tests/integration/test_charm_ambient.py
```diff
@@ -96,6 +109,45 @@
...
+        deadline = time.monotonic() + 60
+        while time.monotonic() < deadline:
+            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as probe:
+                probe.settimeout(1)
+                if probe.connect_ex(("localhost", int(self._port))) == 0:
+                    return f"http://localhost:{self._port}"
+            time.sleep(0.2)
+        raise TimeoutError(
+            f"port-forward to svc/{self._charm_name}:{self._port} was not ready in time"
+        )
```
NohaIhab: why not tenacity? we usually use it, you can define a helper like:
```
@retry(stop=stop_after_delay(60), wait=wait_fixed(0.2), reraise=True)
def _wait_until_ready(self) -> None:
    """Block until the forwarded local port accepts connections."""
    with socket.create_connection(("localhost", int(self._port)), timeout=1):
        pass
```

src/mlflow_client_reconcile.py
MattiaSarti: No, sorry, maybe I didn't explain it clearly enough: the problem is the fact that we grant super-admin privileges to other users (in addition to the charm's) and that they can therefore (intentionally or inadvertently) delete the charm's own admin, so that the charm wouldn't be able to authenticate via HTTP anymore, but if we instead run this code snippet in the workload, the charm can still reconcile any grants, not just super-admin grants, or relation changes, even if its super-admin is gone

metadata.yaml
MattiaSarti: Okay, is [this](https://github.com/canonical/mlflow-operator/pull/494/changes/bcf873bf4d660d851c56d5b7e4c1140325e84b6c) enough?

src/mlflow_client_reconcile.py
NohaIhab: suggestion(non-blocking): can we have unit tests covering the script? right now the script's own logic is only exercised by the integration tests. We can have a separate test file `test_mlflow_client_reconcile.py`. For the store, we can have a minimal in-memory SQLite DB, or whichever way you prefer to _unit_ test it (non-expensive/time-consuming tests).

src/mlflow_client_reconcile.py
NohaIhab: I agree that the workload-side code is needed as a recovery path. My question is whether we can keep the private API usage narrowly separated: use it to restore the charm’s own user and super-admin role when necessary, then use the REST APIs for ordinary user, workspace, and relation reconciliation once authentication is available again.

If the recovery path must also reconcile all grants because the charm may be locked out during a relation change, then I understand why the private store access needs to cover the full desired state. In that case, it would be useful to document it explicitly as a recovery reconciliation mechanism rather than a general replacement for the REST APIs.

tests/integration/test_charm_object_storage.py
NohaIhab: question: are we keeping minio for MLflow 3? just for me to have context on when we are deprecating it

NohaIhab (commented):
Thanks for addressing previous comments, here's some follow-up and non-blocking asks

metadata.yaml
NohaIhab: yes, thx

tests/integration/test_charm_object_storage.py
MattiaSarti: Sure. No, we'll get rid of that before releasing to `stable` risk, it's one of the things we'll implement soon.

src/mlflow_client_reconcile.py
MattiaSarti: > If the recovery path must also reconcile all grants because the charm may be locked out during a relation change, then I understand why the private store access needs to cover the full desired state.

Yes, this is the case.

> In that case, it would be useful to document it explicitly as a recovery reconciliation mechanism rather than a general replacement for the REST APIs.

Okay, I've added a paragraph to the docstring of the script to highlight it.

src/mlflow_client_reconcile.py
MattiaSarti: One one hand, it does make sense to add extra unit tests to foster coverage. On the other hand, even if minimal, they would require considerable extra effort and would make this task spill into the next pulse. If we want to deliver multi-tenancy as soon as possible, we could alternatively park this as a technical improvement for the future, given that we already have those integration tests that cover all use cases. Let's ask @deusebio how to go about this.

tests/integration/test_charm_ambient.py
```diff
@@ -96,6 +110,49 @@
 )
 
 
+class _PortForward:
```
deusebio: **praise** That's a nice implementation! Well done! I like this context!

src/charm.py
```diff
@@ -1157,6 +1183,173 @@ def _reconcile_auth_config(self) -> None:
...
+        or deleted (intentionally or inadvertently) by an external super-admin.
+        """
+        payload = json.dumps(
+            {
+                "users": users_grants_across_workspaces,
+                "protected_admin": MLFLOW_SUPER_ADMIN_USERNAME,
+            }
+        )
+        reconcile_script = Path(MLFLOW_CLIENT_RECONCILE_SOURCE_PATH).read_text()
+        process = self.container.exec(
```
deusebio: **question,non-blocking** I'm not sure why we are running this in the workload container. I mean: the only reason I could see (but correct me if I'm wrong) was to read the config to retrieve the store endpoints and connection parameters. But maybe one could do the other way around, fetch the config file from the workload and then use it from the charm container. 

It just feels weird to have a python script to be copied to the workload, to then be run there. Not a biggie though. It is not that I have really strong concerns here

src/mlflow_client_reconcile.py
deusebio: I don't think it is super critical, this code is tested in integration tests. I raised elsewhere that I feel it is weird to always copy this file to the workload. In this case, it would feel that this actually belongs to the workload. If that's the case, we could probably move this in the image and integration test it there just using docker (if it is so simple that by just running MLFlow we can easily test this script)

Let's make it a technical improvement, also depending whether this will be moved to the rock, left here or to the charm container

deusebio (approved):
Looks good! I only have some small, non-blocking concern about the script that is here but it is run in the workload. Seems strange, a bit more brittle, but deinitely not as critical to block this PR

src/charm.py
MattiaSarti: Sure. The reason we run that in the workload is not related to reading configs (which we can always read or pass in alternative ways), but to directly access MLflow's internals to reconcile grants even if the charm's super-admin is compromised - you can see the comment thread we've had above: https://github.com/canonical/mlflow-operator/pull/494#discussion_r4092060715. So it's not even parametrized at all by the charm and it could also be moved to the rock and tested thoroughly there, yes. Shall we also treat this as a technical improvement?

Issue #491, opened 2026-09-10
AuthorizationPolicy for Multi-Tenancy


MattiaSarti:
### Context

The charm has to support multi-tenancy according to the design outlined in the official spec.

### What needs to get done

Modify the charm logic to operate the required AuthorizationPolicy to allow traffic only:
- to the tracking server when from the platform-namespace waypoint proxy or from either of the two ingress gateways
- to the metrics exporter when from OpenTelemetry collector

### Definition of Done

The charm logic is modified to operate the required AuthorizationPolicy to allow traffic only:
- to the tracking server when from the platform-namespace waypoint proxy and from the two ingress gateways
- to the metrics exporter when from OpenTelemetry collector

NohaIhab:
## Testing
* red path: random pod (not in the mesh) tries to connect and see the connection rejected
* green path: existing tests are enough

NohaIhab:
Closed by #495

PR #495, opened 2026-09-18
feat: update authpolicies for multi-tenancy

NohaIhab:
Closes #491 

## Summary

Add Istio authorization policies for MLflow 3 multi-tenancy and restrict workload access to the following Identities:
* the Waypoint in the platform namespace
* the Ingress gateway(s)
* the Unit related on the `metrics-endpoint` relation (only for the metrics port)

## Changes

- Add the `istio_waypoint_principal` charm configuration option for the platform waypoint SPIFFE principal.
- Create a tracking-server `AuthorizationPolicy` that:
  - selects MLflow backend pods,
  - allows the configured waypoint principal,
  - restricts access to the tracking-server port,
  - is omitted when no waypoint principal is configured.
- Use `UnitPolicy` for the `metrics-endpoint` relation so the OpenTelemetry metrics consumer can access the Prometheus exporter port.
- Rename the managed policy scope to `mlflow-server-tracking-server-access`.
- Add ambient integration coverage proving that a pod outside the mesh is rejected by Istio. The test verifies the observed ztunnel behavior: curl exits with code `56`

### Additional changes
- Replace the charm libraries from `istio_beacon_k8s` and `istio_ingress_k8s` to the new charmlibs.
- Replace `charmed-service-mesh-helpers` with the new `canonical-service-mesh` package


src/charm.py
```diff
@@ -249,12 +257,17 @@ def __init__(self, *args):
 
         # for an ambient-mode service mesh:
 
-        self._mesh = ServiceMeshConsumer(self)
-
-        # Allow all policy needed to allow requests from all user namespaces
-        self._allow_all_policy = generate_allow_all_authorization_policy(
```
NohaIhab, line 255: note that removing this one means we have to clean it up in case of upgrading from MLflow 2.x @MattiaSarti

config.yaml
```diff
@@ -54,3 +54,13 @@ options:
...
+  istio_waypoint_principal:
+    description: |
+      The SPIFFE principal of the platform namespace's waypoint proxy, e.g.
+      "cluster.local/ns/<namespace>/sa/<model>-istio-beacon-k8s-waypoint". This allows in-mesh
+      clients that reach the MLflow tracking server through the waypoint.
+
+      Only relevant when the charm is related over `service-mesh`. When left empty, the charm does
+      not create an authorization policy for the tracking server.
+    type: string
+    default: ""
```
NohaIhab, line 66: should we set a default for this one?

src/charm.py
MattiaSarti: True, but... Charmed MLflow 3 will be backward incompatible with respect to Charmed MLflow 2, so we don't expect users to Juju-refres their charm and we can safely assume they will start from a clean cluster state

config.yaml
MattiaSarti: Good point, but it's hard to guess both the namespace and the model name with a default value... So it's probably safer to leave that empty and have the charm cleanly blocked on that, telling the user it has to be set, rather than having a non-sensible default that would oftentimes fail

src/charm.py
NohaIhab: gotcha, but would there be no upgrade guide?

config.yaml
NohaIhab: https://github.com/canonical/mlflow-operator/pull/495/commits/2ee9302b7934c94f329a296ad242fb63b6284105

src/charm.py
MattiaSarti: No, but we were thinking of providing users with a data-migration guide, even if it's quite tricky because not only they would have to do that themselves from an old charmed database to a new one, but most importantly because that also switched from the old MySQL provider to the new PostgreSQL one, so we're not sure yet - in any case, nothing beyond data migration

config.yaml
MattiaSarti: nitpick: do we want to mention the new behavior in the config's description? Otherwise users may understad the charm will be active and fine when empty with no AuthorizationPolicy

src/charm.py
```diff
@@ -1130,10 +1161,55 @@ def _reconcile_policy_resource_manager(self):
...
             return
         if self.model.get_relation(SERVICE_MESH_RELATION_NAME):
+            # this only manages the waypoint (in-mesh) allowance for the tracking server; each
+            # related ingress's gateway is allowed to reach the tracking server port by a separate
+            # AuthorizationPolicy that the istio-ingress-k8s charm creates automatically, one per
+            # related ingress.
+            tracking_server_policy = self._build_tracking_server_authorization_policy()
             self._policy_resource_manager.reconcile(
-                policies=[], mesh_type=self._mesh.mesh_type, raw_policies=[self._allow_all_policy]
+                policies=[],
```
MattiaSarti, line 1170: question: I'm just trying to understand and I may be completely off track, but should't we include the same `UnitPolicy` as when we initialize `MeshType` here, or it not to disappear? I can't figure it out from [the upstream source](https://github.com/canonical/service-mesh/blob/main/canonical_service_mesh/src/canonical_service_mesh/k8s/resource_manager/_resource_manager.py#L400)

src/charm.py
```diff
@@ -249,12 +262,17 @@ def __init__(self, *args):
...
+        self._mesh = ServiceMeshConsumer(
+            self,
+            policies=[
+                # allow the related metrics consumer (e.g. the OpenTelemetry collector) to reach
+                # the Prometheus exporter port; the mesh generates the authorization policy from
+                # this relation:
+                UnitPolicy(
+                    relation=METRICS_RELATION_NAME,
+                    ports=[int(self._exporter_port)],
+                ),
```
MattiaSarti, line 274: question: an `AppPolicy` with [a single `Endpoint` and the port specified there](https://github.com/canonical/charmlibs/blob/main/interfaces/service_mesh/src/charmlibs/interfaces/service_mesh/_service_mesh.py#L76) would also work, right? How do we choose between these equivalent options? Perhaps it's better to stick with a `UnitPolicy` because OpenTelemetry collector accesses K8s Pods directly instead of via their applications' K8s Services?

src/charm.py
```diff
@@ -1130,10 +1161,55 @@ def _reconcile_policy_resource_manager(self):
...
+        return principal.strip() or None
+
+    def _build_tracking_server_authorization_policy(self) -> Optional[AuthorizationPolicy]:
+        """Build the authorization policy allowing in-mesh access to the tracking server port.
+
+        Only the platform waypoint proxy's principal is configured here: it is the identity the
+        tracking server's ztunnel sees for in-mesh traffic. Ingress traffic is allowed separately,
+        by an AuthorizationPolicy that the istio-ingress-k8s charm creates automatically for each
+        related ingress (permitting that gateway's workload to reach the tracking server port).
+        Returns None when no waypoint principal is configured, so no policy is created.
```
MattiaSarti, line 1187: praise: helpful

tests/integration/test_charm_ambient.py
MattiaSarti, line 1: todo: we should copy the same changes to `test_charm_s3.py`, which is also based on ambient mode, since we're still following the patter of replicating across test files.

tests/integration/test_charm_ambient.py
```diff
@@ -950,3 +991,87 @@ async def test_client_logs_and_fetches_artifact_via_tracking_server(self, ops_te
...
+
+        # Retry to let the authorization policy reach the ztunnel and to tolerate pod startup.
+        for attempt in Retrying(stop=stop_after_delay(300), wait=wait_fixed(15), reraise=True):
+            with attempt:
+                exit_code, stderr = self._curl_tracking_server_from_pod(
+                    out_of_mesh_namespace, tracking_url
+                )
+                # the ztunnel resets the denied L4 connection, so curl exits 56 ("Recv failure:
+                # Connection reset by peer"); anything else (0 = reached the server, or a
+                # pod/kubectl error) is retried until the policy is enforced.
```
MattiaSarti, line 1073: question: does this `Recv failure: Connection reset by peer` give us the guarantee to come exclusively from an Isito denial, though?

src/charm.py
NohaIhab: this only affects policies managed by the prm, which the `UnitPolicy` one isn't, I'll make sure it's covered in unit tests

src/charm.py
NohaIhab: an `AppPolicy` would allow access to the service rather than the pod, and exactly what you said is correct otelcollector needs to access the pod directly

tests/integration/test_charm_ambient.py
NohaIhab: Not necessarily, but it is the expected denial, I tried it in a deployment and this is the exact message and code I get. If we want to be more specific I think we'd have to check the authpolicy for the expected waypoint principal, or inspect ztunnel logs for the denied connection. Do you think this is necessary?

tests/integration/test_charm_ambient.py
NohaIhab: https://github.com/canonical/mlflow-operator/pull/495/commits/5b203e1f7b0edde8ecb00266b8ec45cad3f24cc3

src/charm.py
NohaIhab: https://github.com/canonical/mlflow-operator/pull/495/commits/93819513dfc9c17df81f0bbe2e594119c62f6437

tests/integration/test_charm_ambient.py
MattiaSarti: Got it, then no need, I think it's enough

MattiaSarti (approved):

config.yaml
NohaIhab: https://github.com/canonical/mlflow-operator/pull/495/commits/9b8266ad7619e0f26db47714dcfa416085fde899

misohu (approved):

